PR #12, opened 2019-06-11
Fix #11: Allow to restore good value of topology

### LINKED ISSUE #11: Need to correct topology section with old good value in case of unauthorized change
when asking a non-authorized scaledown to 0 the Operator won't execute the action but should also correct the value in the CRD definition to it's old value.

This is not actually the case

The log message are good : 

```
time="2019-06-11T07:54:49Z" level=info msg="Ask ScaleDown to 0 for dc dc3" cluster=cassandra-demo
time="2019-06-11T07:54:50Z" level=warning msg="The Operator has refused the ScaleDown. Keyspaces still having data [system_distributed bench system_auth system_traces]" cluster=cassandra-demo
```

but the value is not corrected in topology section

allamand:
Restore good value when CassKop refuse to apply a change.
We used a local var instead of the global needUpdate one.

### DIFF
--- a/pkg/controller/cassandracluster/reconcile.go
+++ b/pkg/controller/cassandracluster/reconcile.go
@@ -120,14 +120,15 @@ func (rcc *ReconcileCassandraCluster) CheckNonAllowedChanged(cc *api.CassandraCl
 		return true
 	}
 
-	if needUpdate, updateStatus := CheckNonAllowedRemoveDC(rcc, cc, status, &oldCRD); needUpdate {
+	var updateStatus string
+	if needUpdate, updateStatus = CheckNonAllowedRemoveDC(rcc, cc, status, &oldCRD); needUpdate {
 		if updateStatus != "" {
 			status.LastClusterAction = updateStatus
 		}
 		return true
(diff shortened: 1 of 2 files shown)


pkg/controller/cassandracluster/reconcile.go
```diff
@@ -120,14 +120,15 @@ func (rcc *ReconcileCassandraCluster) CheckNonAllowedChanged(cc *api.CassandraCl
 		return true
 	}
 
-	if needUpdate, updateStatus := CheckNonAllowedRemoveDC(rcc, cc, status, &oldCRD); needUpdate {
+	var updateStatus string
```
Orange-cscetbon: @allamand, I can't see why this change is needed ? In all cases we overwrite those 2 variables so why not just keep the old code ?

allamand: @Orange-cscetbon with the previous change we uses := which in fact creates 2 variables.
so the needUpdate var in this case was not the global needUpdate one which leads to the bug encountered

PR #14, opened 2019-06-11
Update CircleCI configuration

allamand:
so that we can execute build on PR without sharing environment secrets


allamand:
close

Issue #17, opened 2019-06-13
Helm chart tags problems


hkroger:
The "label" key is twice and labels are indented wrong here: https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/1f38e76fa377f63071b33748628a768795854398/helm/cassandra-operator/templates/service.yaml#L8

I think here the indentation is wrong (4 spaces instead of 2) for labels: https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/1f38e76fa377f63071b33748628a768795854398/helm/cassandra-operator/templates/rolebinding.yaml#L6

I think this should also have labels (if I install operator with helm chart, uninstall it and install again the CRD given an error because it already exists): https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/1f38e76fa377f63071b33748628a768795854398/helm/cassandra-operator/templates/db_v1alpha1_cassandracluster_crd.yaml#L3

allamand:
Hi @hkroger,

for the crd problem on reinstall this must be manage via the --no-hooks flag cf https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/master/helm/cassandra-operator/readme.md#install-of-the-crd

PR #18, opened 2019-06-13
Fix helm chart labels

hkroger:
Fix problems related to #17.

Removed 'crd-install' hook because CRDs are installed anyways before other components. See: https://stackoverflow.com/questions/51957676/helm-install-in-certain-order

'crd-install' causes CRD not to be uninstalled by default when release is deleted.

What do you think?

### DIFF
--- a/helm/cassandra-operator/Chart.yaml
+++ b/helm/cassandra-operator/Chart.yaml
@@ -5,7 +5,7 @@ home: https://github.com/Orange-OpenSource/cassandra-k8s-operator
 sources:
   - https://github.com/Orange-OpenSource/cassandra-k8s-operator
 version: 0.3.1
-appVersion: 0.3.1-mater
+appVersion: 0.3.1-master
 icon: 
 maintainers:
   - name: allamand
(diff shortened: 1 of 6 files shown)


allamand:
Hello @hkroger,

thanks for contributing this is awsome :)

Concerning the crd-install If you remove it, then the helm won't install it and the operator wont works.
In fact this is needed to be installed through helm in order to pass the automatic tests on charts/helm : https://github.com/helm/charts/pull/14414

Also the fact that if we delete the Helm, the CRD is not deleted is actually a "feature" because if we delete the CRD, then all associated ressources (cassandra clusters) are also deleted and this is not what we wanted in the first place.

allamand:
For the rest I agree but for reinstall the crd with helm you can disable the hook with --no-hooks cf https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/master/helm/cassandra-operator/readme.md#install-of-the-crd.

Can you resend the PR without removing the crd-hook ?

thanks

hkroger:
@allamand, alright if you prefer it this way. Not a problem for me :) I pushed an updated version.

allamand:
@hkroger thanks a lot for you're contribution. I'll merge it now :)

PR #19, opened 2019-06-15
add step for adding help incubator repo

jsanda:
For users who are not familiar or experienced with Helm, it is not clear from the docs how to access and use the charts from the incubator repo. This PR adds a step for adding the incubator repo.


cscetbon:
It's definitely needed. Thanks for contributing !

PR #22, opened 2019-06-19
Set createCustomResource: true to default deploying the CRD 

ahmedjami:
This will allow the creation of CRD at the step of deploying casskop through helm.


allamand:
Thanks ;)

PR #31, opened 2019-06-27
add a note about creating the cassandra-e2e namespace

jsanda:
I am not sure how things are set up for CI, but for running e2e tests locally I had to first create the `cassandra-e2e` namespace. It would be helpful to have that explicitly stated in the developer docs.


allamand:
Thanks ;)

Issue #41, opened 2019-07-05
Update the service


allamand:
Actually, we have 1 service for each dc defined in the CassandraCluster.

We want to change to have only one service for the entire cluster.

Application clients will needs do specify on which Cassandra DC they want to connect when establishing the connection.

jsanda:
@allamand I would be happy to work on this. I can spend some time on it over the weekend.

allamand:
yes thanks :)

jsanda:
@allamand should there be a single monitoring service for the entire cluster as well?

allamand:
@jsanda yes for monitoring too

jsanda:
> Application clients will needs do specify on which Cassandra DC they want to connect when establishing the connection.

This would be done by using a `DCAwareRoundRobinPolicy` for the `LoadBalaningPolicy` right?

jsanda:
Are there are any other changes needed for this PR? If not, do you want me to squash the commits?

jsanda:
@allamand can this be closed?

PR #42, opened 2019-07-09
Create a single service for the whole cluster, not per DC

jsanda:
This PR is for #41.I still need to add some automated tests and possibly change the monitoring service as well.

### DIFF
--- a/test/e2e/cassandracluster_test.go
+++ b/test/e2e/cassandracluster_test.go
@@ -242,10 +242,9 @@ func cassandraClusterServiceTest(t *testing.T, f *framework.Framework, ctx *fram
 	assert.Equal(t, kind, clusterService.ObjectMeta.OwnerReferences[0].Kind)
 	assert.Equal(t, cluster.Name, clusterService.ObjectMeta.OwnerReferences[0].Name)
 	assert.True(t, *clusterService.ObjectMeta.OwnerReferences[0].Controller)
-	assert.Equal(t, 3, len(clusterService.Spec.Ports))
+	assert.Equal(t, 4, len(clusterService.Spec.Ports))
 
-	assertServiceExposesPort(t, &clusterService, "cassandra-port", 9042)
-	assertServiceExposesPort(t, &clusterService, "http-metrics", 9121)
+	assertServiceExposesPort(t, &clusterService, "cql", 9042)
 	assertServiceExposesPort(t, &clusterService, "intra-node", 7000)
 	assertServiceExposesPort(t, &clusterService, "intra-node-tls", 7001)
 	assertServiceExposesPort(t, &clusterService, "jmx-port", 7199)
@@ -256,7 +255,7 @@ func cassandraClusterServiceTest(t *testing.T, f *framework.Framework, ctx *fram
 	assert.True(t, *monitoringService.ObjectMeta.OwnerReferences[0].Controller)
 	assert.Equal(t, 1, len(monitoringService.Spec.Ports))
 
-	assertServiceExposesPort(t, &monitoringService, "http-promcassjmx", 1234)
+	assertServiceExposesPort(t, &monitoringService, "promjmx", 1234)
 }
 
 func NoCleanup() *framework.CleanupOptions {
--- a/test/e2e/cassandracluster_test.go
+++ b/test/e2e/cassandracluster_test.go
@@ -17,7 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
-	log "github.com/sirupsen/logrus"
+	logrus "github.com/sirupsen/logrus"
 )
 
 // Run all fonctional tests
@@ -205,7 +205,7 @@ func cassandraClusterServiceTest(t *testing.T, f *framework.Framework, ctx *fram
 		},
 	}
 
-	log.Debugf("Creating cluster")
+	logrus.Debugf("Creating cluster")
 	if err := f.Client.Create(goctx.TODO(), cluster, NoCleanup()); err != nil && !apierrors.IsAlreadyExists(err) {
 		t.Fatalf("Error Creating CassandraCluster: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


test/e2e/cassandracluster_test.go
```diff
@@ -144,3 +150,124 @@ func cassandraClusterRollingRestartDCTest(t *testing.T, f *framework.Framework,
...
+			NodesPerRacks: 2,
+			HardAntiAffinity: false,
+			DeletePVC: true,
+			Resources: api.CassandraResources{
+				Limits: api.CPUAndMem{
+					CPU: "500m",
+					Memory: "1Gi",
+				},
+			},
+			Topology: api.Topology{
```
Orange-cscetbon: Please try to use the minimum required for your tests in order to minimize the time e2e tests take.
For example, using 1 node per rack and only one rack per DC here.

test/e2e/cassandracluster_test.go
```diff
@@ -144,3 +150,124 @@ func cassandraClusterRollingRestartDCTest(t *testing.T, f *framework.Framework,
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "multidc-cluster",
+			Namespace: namespace,
+			Labels:    map[string]string{"cluster": "k8s.pic"},
+		},
+		Spec: api.CassandraClusterSpec{
+			BaseImage: "orangeopensource/cassandra-image",
+			NodesPerRacks: 2,
+			HardAntiAffinity: false,
```
Orange-cscetbon: You can remove it, it's false by default

test/e2e/cassandracluster_test.go
```diff
@@ -41,7 +47,7 @@ func CassandraClusterTest(code func(t *testing.T, f *framework.Framework,
 	ctx *framework.TestCtx)) func(t *testing.T) {
 	return func(t *testing.T) {
 		ctx, f := mye2eutil.HelperInitOperator(t)
-		defer ctx.Cleanup()
+		//defer ctx.Cleanup()
```
Orange-cscetbon: any reason why it's commented ?

test/e2e/cassandracluster_test.go
```diff
@@ -144,3 +150,124 @@ func cassandraClusterRollingRestartDCTest(t *testing.T, f *framework.Framework,
...
+	assert.Equal(t, 2, len(services.Items))
+
+	// TODO add verification of services as well as statefulsets, etc.
+}
+
+func cleanupOptions(ctx *framework.TestCtx) *framework.CleanupOptions {
+	return NoCleanup()
+}
+
+func NoCleanup() *framework.CleanupOptions {
```
Orange-cscetbon: It seems you only call that function in `cleanupOptions`, then why do you need it ?

test/e2e/cassandracluster_test.go
jsanda: Because the e2e tests are expensive to run, I kind of felt like it would be overkill to have the test only verify that a single service is created for the cluster. I figured I might as well verify that other objects, namely StatefulSet, get created. 

While I think what I described would be a good test to have, I am fine with using the minimum resources necessary and leaving a multi-DC/multi-rack scenario for a separate test. Would that be better?

test/e2e/cassandracluster_test.go
jsanda: 👍

test/e2e/cassandracluster_test.go
jsanda: This is just debugging. I am having trouble figuring out how to have resources not get deleted when tests finish running. I will revert that code.

test/e2e/cassandracluster_test.go
jsanda: I added that in anticipation of needing `*framework.CleanupOptions` in multiple places. I wanted to have a single function that controls whether cleanup is performed for debugging; however, as I mentioned in my comment above, I am having trouble figuring out how to prevent resources from getting deleted after tests run.

jsanda:
@Orange-cscetbon and/or @allamand can you please review?

I have already pushed changes to address the earlier comments.

Thanks

John

allamand:
@jsanda, lgtm, you may need to add your test in the CircleCI pipeline to be automatically tested https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/master/.circleci/config.yml#L286

jsanda:
>you may need to add your test in the CircleCI pipeline to be automatically tested 

done!

test/e2e/cassandracluster_test.go
cscetbon: Yeah I think so. Tests are gonna take longer the more we add in here. Trying to limit the number of times we need to instantiate a cluster (which takes time) makes more sense to me.

test/e2e/cassandracluster_test.go
cscetbon: oh I see, so your PR is not ready yet for complete review. You should label it with `wip` then

test/e2e/cassandracluster_test.go
```diff
@@ -144,3 +152,179 @@ func cassandraClusterRollingRestartDCTest(t *testing.T, f *framework.Framework,
...
+	}
+
+	assert.Equal(t, 1, len(clusterService.ObjectMeta.OwnerReferences))
+	assert.Equal(t, kind, clusterService.ObjectMeta.OwnerReferences[0].Kind)
+	assert.Equal(t, cluster.Name, clusterService.ObjectMeta.OwnerReferences[0].Name)
+	assert.True(t, *clusterService.ObjectMeta.OwnerReferences[0].Controller)
+	assert.Equal(t, 3, len(clusterService.Spec.Ports))
+
+	assertServiceExposesPort(t, &clusterService, "cassandra-port", 9042)
+	assertServiceExposesPort(t, &clusterService, "cassandra-thrift", 9160)
```
cscetbon: I still think we should stop supporting thrift. what do you think @jal06, @allamand ?

allamand: I have remove thrift from CassKop in a separate PR. we should also remove it from tests

allamand:
@jsanda I've made some changes in master in order to be able to tests from PRs.
This takes at this time a manual steps on our side to trigger the tests.
Can you rebase on master ?

jsanda:
Are there any other changes needed for this PR? If not do you want me to squash the commits?

cscetbon:
@allamand where are we on this ?

allamand:
Thanks @jsanda for the good work :)

PR #54, opened 2019-07-14
volume-permissions - volumes are rw by default

cscetbon:


cscetbon:
@allamand can you take a look ? there is no reason tests fail because of that change

allamand:
needs to be rebased on master for triggering tests

PR #57, opened 2019-07-15
fix the bug that error is not handled

mackwong:
- Fix problems that error is not handled
- No need to judge whether `labels` is nil

### DIFF
--- a/pkg/controller/cassandracluster/statefulset.go
+++ b/pkg/controller/cassandracluster/statefulset.go
@@ -95,6 +95,9 @@ func (rcc *ReconcileCassandraCluster) UpdateStatefulSet(statefulSet *appsv1.Stat
 	//Check that the new revision of statefulset has been taken into account
 	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
 		newSts, err := rcc.GetStatefulSet(statefulSet.Namespace, statefulSet.Name)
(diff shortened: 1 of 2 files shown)


pkg/controller/cassandracluster/statefulset.go
```diff
@@ -95,6 +95,9 @@ func (rcc *ReconcileCassandraCluster) UpdateStatefulSet(statefulSet *appsv1.Stat
 	//Check that the new revision of statefulset has been taken into account
 	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
 		newSts, err := rcc.GetStatefulSet(statefulSet.Namespace, statefulSet.Name)
+		if err != nil && !apierrors.IsNotFound(err){
+			return false, fmt.Errorf("failed to get cassandra statefulset: %cc", err)
```
cscetbon: AFAIK all our logs start with an uppercase. It should be the same here

mackwong: I think it should be as same as https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/4f837cc72be55a0814861a096606403afe9985fb/pkg/controller/cassandracluster/statefulset.go#L93

allamand:
Hi @mackwong, why did you close that one ?

allamand:
ah ok you recreate at #60

PR #60, opened 2019-07-16
fix error is not handled

mackwong:
- Fix problems that error is not handled
- No need to judge whether labels is nil


allamand:
I've triggered tests, /lgtm

allamand:
Thanks @mackwong

PR #63, opened 2019-07-17
jmx exporter name  not match with the one in readme

mackwong:
the `endpoints.port: http-promcassjmx ` is not match with `promjmx`:
https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/8e5f0482d7419124986f774e5bfc0077b73456b4/pkg/controller/cassandracluster/deploy_cassandra.go#L43

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -348,7 +348,7 @@ spec:
       - cassandra-demo
       - default
   endpoints:
-  - port: http-promcassjmx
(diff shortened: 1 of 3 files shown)


allamand:
Hi @mackwong, thanks you for this. I renamed thoses lately and forget to update the docs

Issue #67, opened 2019-07-19
Set spec.resources.limits to spec.resources.requests when not defined


cscetbon:
If **spec.resources.limits** is not set, cassandra pods can consume far more resources than expected and cause issues. It's easy to test it locally in a dind cluster and see that when a second node joins CPU sky rocks. The goal of this ticket is to set it to **spec.resources.requests** if it's not defined to avoid such a situation.

jsanda:
What if `spec.resources.requests` is not set? I am currently using a GKE cluster and there is a LimitRange that specifies defaults for cpu, but there are no defaults for memory. I need to do some more reading, but I am inclined to say that in absence of anything else, memory defaults will come from cgroups.

Orange-cscetbon:
I don't think we should ever accept a`CassandraCluster` object where spec.resources.requests is not set. @allamand wdyt ? If it's not the case we should definitely add it in the current code change I'm asking.

jsanda:
I will work on this. Can someone assign it to me?

cscetbon:
@jsanda let us know if you need any guiding.

jsanda:
@cscetbon can this be closed?

Issue #34, opened 2019-06-28
add structural OpenAPI v3.0 validation schema to CassandraCluster CRD


allamand:
CustomResources traditionally store arbitrary JSON (next to apiVersion, kind and metadata, which is validated by the API server implicitly). With OpenAPI v3.0 validation a schema can be specified, which is validated during creation and updates

With apiextensions.k8s.io/v1 the definition of a structural schema will be mandatory for CustomResourceDefinitions

reference : https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

jsanda:
I can spend some time on this over the weekend.

cscetbon:
It’s be great. Thanks @jsanda

jsanda:
@cscetbon can this be closed?

cscetbon:
Yes closing it, thanks

PR #72, opened 2019-07-21
[WIP] default resource limits to resource requests when limits not set

jsanda:
This is for #67.

### DIFF
--- a/pkg/apis/db/v1alpha1/cassandracluster_types_test.go
+++ b/pkg/apis/db/v1alpha1/cassandracluster_types_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func helperLoadBytes(t *testing.T, name string) []byte {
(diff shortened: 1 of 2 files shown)


pkg/apis/db/v1alpha1/cassandracluster_types_test.go
```diff
@@ -457,3 +458,45 @@ func TestComputeLastAppliedConfiguration(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	cluster.SetDefaults()
+
+	assert.Equal("orangeopensource/cassandra-image", cluster.Spec.BaseImage)
+	assert.Equal(CPUAndMem{CPU: "500m", Memory: "1Gi"}, cluster.Spec.Resources.Limits)
+}
```
cscetbon: Let’s add tests for the other default parameters that are set, wdyt ?

jsanda: >Let’s add tests for the other default parameters that are set, wdyt ?

👍 Absolutely!

I actually spent more time looking into validation, but will that be addressed with #34?

jsanda: @cscetbon I update `TestSetDefaults`.  Do you want to address validation in a separate PR for #34? If so, is there anything else needed for this PR?

cscetbon: Yes, no need to spend time here on validation. Thanks for your contribution !

Orange-cscetbon:
@allamand I approved that PR. Please, do your magic for the e2e tests as it prevents me from merging ...

PR #73, opened 2019-07-23
(Un)Pause operations

cscetbon:
- Add 2 new options to pause/unpause operations
- Apply fields and labels selectors when looking for pods
- Update all labels of a pod at once

See https://pastebin.com/cZGPrpq3 for test cases

### DIFF
--- a/plugins/kubectl-casskop
+++ b/plugins/kubectl-casskop
@@ -21,6 +21,9 @@ from subprocess import check_output, STDOUT, CalledProcessError
 from os.path import basename
 from random import shuffle
 
+PAUSE_LABELS_OP = ["ToDo", "Paused"]
+NO_PODS_FOUND = "No pods found for operation"
+
 def k_apply_with_input(input, error, *args):
     try:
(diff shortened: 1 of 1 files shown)


plugins/kubectl-casskop
```diff
@@ -20,6 +20,11 @@ import json
 from subprocess import check_output, STDOUT, CalledProcessError
 from os.path import basename
 from random import shuffle
+from itertools import groupby
+
+PAUSE_LABELS_OP = ["ToDo", "Paused"]
+NO_PODS_FOUND = "No pods found for operation"
+NO_ONGOING_OP = "--selector=operation-status notin (Ongoing, Finalizing)"
```
allamand: not in

cscetbon: It’s not python here ^ but a selector filter expression.
The syntax is `notin` and not `not in` 🤷‍♂️

cscetbon:
> maybe this is out of scope of this PR but seams to be addressed

Agreed. Let’s create an issue to add tests using the plugin.

plugins/kubectl-casskop
allamand: ok :)

PR #80, opened 2019-07-24
exclude generated files from sonar

allamand:


allamand:
This was an old branch, I forgot to makes the PR.
the goal it to not check generated files in sonar

PR #94, opened 2019-08-16
Enable dynamic snitch

cscetbon:
Dynamic snitch is needed to set dynamic_snitch_badness_threshold in case of a rebuild to avoid the node to be queried by customers

### DIFF
--- a/samples/cassandra-configmap-v1.yaml
+++ b/samples/cassandra-configmap-v1.yaml
@@ -835,7 +835,7 @@ data:
     # of the snitch, which will be assumed to be on your classpath.
     endpoint_snitch: SimpleSnitch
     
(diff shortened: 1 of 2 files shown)


jsanda:
https://issues.apache.org/jira/browse/CASSANDRA-6908 is worth a look. The dynamic snitch often causes performance problems and in general would recommend keeping it disabled.

cscetbon:
@jsanda we saw that ticket and that's why we first disable it by default.

However, it happens only with large clusters and we intend to use dynamic_snitch_badness_threshold to avoid a rebuilding node from getting request from customers and it's a requirement. So we chose to go that way. Anybody that creates his cluster can always override that parameter by disabling it in their custom configmap.

PR #95, opened 2019-08-20
fix hard code in prometheus-grafana-cassandra-dashboard.json

mackwong:
fix hard code in prometheus-grafana-cassandra-dashboard.json


cscetbon:
Thanks @mackwong !

/tests

PR #97, opened 2019-08-21
add statefulset recovery action

mackwong:
I think the reason for https://github.com/Orange-OpenSource/cassandra-k8s-operator/issues/96 is that after deleting the pod, the ```lastClusterAction```  has not been updated,  and ```lastClusterActionStatus``` is always **Done**, so the ```status.Phase``` can not be updated
https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/73fd24e4d80d853a640defa42b91345cb29847ba/pkg/controller/cassandracluster/reconcile.go#L527

### DIFF
--- a/pkg/controller/cassandracluster/cassandra_status.go
+++ b/pkg/controller/cassandracluster/cassandra_status.go
@@ -98,12 +98,17 @@ func (rcc *ReconcileCassandraCluster) getNextCassandraClusterStatus(cc *api.Cass
 	// action.status=Continue (which is set when decommission is successful) will be tested to see if we need to
 	// decommission more
 	// We don't want to check for new operation while there are already ongoing one in order not to break them (ie decommission..)
-    // Meanwhile we allow to check for new changes if _unlockNextOperation	 has been set (to recover from problems)
+	// Meanwhile we allow to check for new changes if _unlockNextOperation	 has been set (to recover from problems)
 	if needSpecificChange ||
 		(!rcc.thereIsPodDisruption() &&
 		  lastAction.Status != api.StatusOngoing &&
 		  lastAction.Status != api.StatusToDo &&
 		  lastAction.Status != api.StatusFinalizing){
+
+		//Update Status if Pod is not Ready
+		if UpdateStatusIfStatefulSetStatusHasChanged(cc, dcRackName, storedStatefulSet, status) {
+			return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/cassandracluster/cassandra_status.go
```diff
@@ -337,6 +356,23 @@ func (rcc *ReconcileCassandraCluster) UpdateStatusIfActionEnded(cc *api.Cassandr
 		nodesPerRacks := cc.GetNodesPerRacks(dcRackName)
 		switch lastAction.Name {
 
+		case api.ActionRecoverStatefulSet:
+
+			if !isStatefulSetNotReady(storedStatefulSet) {
+				logrus.WithFields(logrus.Fields{"cluster": cc.Name, "rack": dcRackName}).Info("Statefulset is Ready")
+				lastAction.Status = api.StatusDone
+				lastAction.EndTime = &now
+				labels := map[string]string{"operation-name": api.OperationCleanup}
```
Orange-cscetbon: I don't understand why there is a cleanup operation in that case.

pkg/controller/cassandracluster/cassandra_status.go
```diff
@@ -182,6 +187,20 @@ func UpdateStatusIfUpdateResources(cc *api.CassandraCluster, dcRackName string,
 }
 */
 
+//UpdateStatusIfStatefulSetStatusHasChanged updates CassandraCluster Action Status if any pod is not Ready
+func UpdateStatusIfStatefulSetStatusHasChanged(cc *api.CassandraCluster, dcRackName string, storedStatefulSet *appsv1.StatefulSet, status *api.CassandraClusterStatus) bool {
```
Orange-cscetbon: This one looks dangerous to me. It can't happen during the creation of the cluster, whenever there is an operation at the stfs level etc..
@allamand don't hesitate to weigh in.

Orange-cscetbon:
I think there is a misunderstanding of what the operator does and the way it checks different status changes or if there are pending operation.

pkg/controller/cassandracluster/cassandra_status.go
mackwong: yes, I didn't take into account the creation of the cluster, 
I will close it.

PR #99, opened 2019-08-27
Prevent rack changes

cscetbon:
This PR introduces the use of https://github.com/r3labs/diff to handle topology changes. A few functions were written to be handle to change changes and be able to check conditions. 
Other unit tests were added to test the function `hasChange` and unit tests were renamed to use Go naming convention.

Those changes were then used to prevent rack changes.

### DIFF
--- a/pkg/controller/cassandracluster/reconcile.go
+++ b/pkg/controller/cassandracluster/reconcile.go
@@ -19,9 +19,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
+	"strings"
 
 	api "github.com/Orange-OpenSource/cassandra-k8s-operator/pkg/apis/db/v1alpha1"
 	"github.com/Orange-OpenSource/cassandra-k8s-operator/pkg/k8s"
+	"github.com/r3labs/diff"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


pkg/controller/cassandracluster/reconcile.go
```diff
@@ -169,113 +182,123 @@ func (rcc *ReconcileCassandraCluster) CheckNonAllowedChanged(cc *api.CassandraCl
...
-			logrus.WithFields(logrus.Fields{"cluster": cc.Name}).Warningf(
-				"The Operator has refused the Topology changed. You can only remove 1 DC at a time, "+
-					"not only a Rack: %v restored to %v", cc.Spec.Topology, oldCRD.Spec.Topology)
-			cc.Spec.Topology = oldCRD.Spec.Topology
-			return true, api.ActionCorrectCRDConfig
+func generatePaths(s string) []string {
+	return strings.Split(s, ".")
+}
+
+func lookForFilter(path []string, filtersRule [][]string, filtersFound *map[string]bool) {
```
allamand: need a comment on the purpose of the function

pkg/controller/cassandracluster/reconcile.go
```diff
@@ -169,113 +182,123 @@ func (rcc *ReconcileCassandraCluster) CheckNonAllowedChanged(cc *api.CassandraCl
...
+				}
+			}
 		}
+	}
+}
 
-		//If we ask to remove only a rack, then it is not authorized
-		if dcsize == olddcsize {
+// hasChange returns if there a the changes with the type provided and only for all paths
+// paths must be provided in ascending order like ('DC', 'DC.Rack') in order to work
```
allamand: in the usage, you uses (.., 'DC.Rack', '-DC') so it's not always the ascending order ?

PR #101, opened 2019-08-27
Remove Cassandra vars from generator script

YuriHeupa:
Remove the Cassandra variables that should be overridden by a configmap from the generator script.

### DIFF
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -45,9 +45,6 @@ const (
 	defaultJvmMaxHeap      = "2048M"
 	hostnameTopologyKey    = "kubernetes.io/hostname"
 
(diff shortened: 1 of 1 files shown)


Orange-cscetbon:
@allamand, @YuriHeupa I don't think adding this variable back is needed. We should remove POD_NAMESPACE from the run.sh script instead. there is no reason to use hostnames with the namespace as all pods are in the same namespace.

I agree regarding the Cassandra variables that it should be overridden by a configmap.

YuriHeupa:
@Orange-cscetbon Makes sense, I'll remove the `POD_NAMESPACE` change from this PR and will keep the Cassandra variables stuff.

YuriHeupa:
@Orange-cscetbon PR updated.

PR #113, opened 2019-09-24
fix nil point bug that caused by scale up datecenter

mackwong:
the pr can fix https://github.com/Orange-OpenSource/cassandra-k8s-operator/issues/108


pkg/controller/cassandracluster/reconcile.go
```diff
@@ -530,7 +529,12 @@ func UpdateCassandraClusterStatusPhase(cc *api.CassandraCluster, status *api.Cas
 
 			rackName := cc.GetRackName(dc, rack)
 			dcRackName := cc.GetDCRackName(dcName, rackName)
-			dcRackStatus := status.CassandraRackStatus[dcRackName]
+			dcRackStatus, ok := status.CassandraRackStatus[dcRackName]
```
cscetbon: You could rename ok and use something like `dcRackExists` which makes more sense

mackwong: have changed it to `exist`

cscetbon:
@allamand I think we can merge that one. The timeout issue is probably fixed in master as you know.

PR #119, opened 2019-10-04
Refactor makefiles and do some cleaning

cscetbon:


.circleci/config.yml
```diff
@@ -221,16 +225,9 @@ jobs:
...
-          -Dsonar.login=${SONAR_LOGIN} \
           -Dsonar.log.level=INFO
           cat -n .scannerwork/report-task.txt
 
-    - save_cache:
-        name: Cache SonarQube Scanner
-        key: sonar-scanner-3.3.0.1492-linux
-        paths:
-          - sonar-scanner-3.3.0.1492-linux
-          - /home/circleci/cassandra-k8s-operator/.sonar
```
allamand: Why move save cache before action ??

.circleci/config.yml
```diff
@@ -134,7 +133,6 @@ jobs:
         paths:
           - build/_output
           - pkg/apis/db/v1alpha1/zz_generated.deepcopy.go
-          - pkg/apis/db/v1alpha1/zz_generated.deepcopy.go
```
allamand: We need to cache the other generated file

docker/circleci/Dockerfile
```diff
@@ -103,12 +87,6 @@ RUN mkdir -p $GOPATH/src/github.com/operator-framework \
...
     && rm -rf /go/pkg/dep/sources/
 
-
-
-# END IMAGE CUSTOMIZATIONS
-
-#USER circleci
-
-WORKDIR cassandra-k8s-operator
+WORKDIR /go/cassandra-k8s-operator
```
allamand: Why adding /go there ?

cscetbon: `Follow best practices`

.circleci/config.yml
cscetbon: just not needed later

cscetbon:
> Are you sure about this ?

Blocked due to free-plan-docker-layer-caching-unavailable

docker/circleci/Dockerfile
allamand: There is no need since we used go modules

cscetbon: Not sure we're on the same page. What do you mean ? I'm just saying that it's a best practice to use absolute paths as workdir values, see https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/

allamand: Just why adding a /go in the path ?

.circleci/config.yml
allamand: Why ?

cscetbon: We don’t need to cache the same file twice and there is only one generated file. Default is not there anymore

docker/circleci/Dockerfile
cscetbon: Read the part on workdir, it must be absolute. And this is where it goes when we don’t specify it. So I’m not changing anything, just applying best practices

Issue #89, opened 2019-08-06
Custom Resource Spec property names and kube-openapi 


devdattakulkarni:
Looking through the Custom Resource type definitions, I found following properties that may not be adhering to kube-openapi requirements.

CassandraClusterSpec.GCStdout
CassandraClusterSpec.UnlockNextOperation
PodLastOperation.Name
CassandraLastAction.Name
CassandraClusterStatus.SeedList
CPUAndMem.CPU
Topology.DC

Kube OpenAPI name validation rules expect the field name in Go code and field name in JSON to be exactly same with just the first letter in different case (Go code requires CamelCase, JSON requires camelCase).

Here are some other best practice guidelines around Operator development, packaging, documentation: https://github.com/cloud-ark/kubeplus/blob/master/Guidelines.md

jsanda:
@devdattakulkarni thanks for point this out. I will address this in #87.

PR #87, opened 2019-08-03
[WIP] initial schema after running `operator-sdk generate openapi`

jsanda:
This PR is for #34. 

**Note:** This is a WIP and not ready to be merged.

I am opening the PR so I can gather feedback as I make changes.

PR #122, opened 2019-10-08
OpenAPI v3.0 schema validation

jsanda:
This PR is for #34.


jsanda:
@allamand / @cscetbon the initial validation changes are in place. I have some questions:

Do we want regex validation for `CassandraResources` for cpu and memory?

Do we want to make sure `DataCapacity` is > 0 (if specified)?

Do we want to make sure `MaxPodUnavailable` is > 0 (if specified)?

Do we want to make sure `ImagePullPolicy` has a valid value (this would be done with an enum tag)?

Do we want to make sure `RunAsUser` is > 0 (if specified)?

Something else that I thought could be interesting is that if `ConfigMapName` is set, then make sure that there is a configmap with the specified name. This would have to be done with a validating admission web hook. If there is interest in something like this, I would be happy to do it in a follow up PR.

jsanda:
@allamand / @cscetbon 

Any thoughts on this? And can you run the e2e tests please?

allamand:
Hi John,

CircleCI has removed docker layer caching for free tiers. We need to update circleCI config to remove that. You can wait for my branch to be merged then rebased on it

allamand:
Hi @jsanda, 
> Do we want regex validation for `CassandraResources` for cpu and memory?
I'm not sure how this is done for pod for instance, we should do same thing here

> Do we want to make sure `DataCapacity` is > 0 (if specified)?
This makes sense
 
> Do we want to make sure `MaxPodUnavailable` is > 0 (if specified)?
I wonder If MaxPodUnavailable=0 could be useful in some cases ?

> Do we want to make sure `ImagePullPolicy` has a valid value (this would be done with an enum tag)?
This makes sense

> Do we want to make sure `RunAsUser` is > 0 (if specified)?
This makes sense

> Something else that I thought could be interesting is that if `ConfigMapName` is set, then make sure that there is a configmap with the specified name. This would have to be done with a validating admission web hook. If there is interest in something like this, I would be happy to do it in a follow up PR.
This could be useful also, but not mandatory (or configurable). Four our case we know we won't be able to install some admission webhook in our multi-tenant, this will be forbidden by our admins.
Saying that, this could be useful for folks that are allowed to install some.

jsanda:
@allamand 

I have added additional validations. Is there anything else that should be done in this PR?

cscetbon:
@allamand, please answer john’s question.

fdehay:
Hi @jsanda, does this PR fix #89 like you said in #87 ? Seems yes but not too sure so I prefer asking :)

jsanda:
Hi @fdehay,

I pushed a commit that fixes the naming for the following:

* `CassandraClusterSpec.UnlockNextOperation`
* `PodLastOperation.Name`
* `CassandraLastAction.Name`
* `CassandraClusterStatus.SeedList`

I have not made changes to:

* `CassandraClusterSpec.GCStdout`
* `CPUAndMem.CPU`
* `Topology.DC`

I did not change these because there are different conventions for how to handle abbreviations and acronyms in camel case. I think that the Go identifier names are fine. I am just now sure what the JSON names should be.

deploy/crds/db_v1alpha1_cassandracluster_crd.yaml
```diff
@@ -10,4 +10,313 @@ spec:
...
+          properties:
+            cassandraRackStatus:
+              additionalProperties:
+                properties:
+                  cassandraLastAction:
+                    description: 'CassandraLastAction is the set of Cassandra State
+                      & Actions: Active, Standby..'
+                    properties:
+                      Name:
+                        description: 'Type d''action a effectuer : UpdateVersion,
```
cscetbon: @jsanda you're a french speaker now ? ha ha. It should be translated

deploy/crds/db_v1alpha1_cassandracluster_crd.yaml
```diff
@@ -10,4 +10,313 @@ spec:
...
+                        description: List of Racks defined in the Cassandra DC
+                        items:
+                          properties:
+                            labels:
+                              additionalProperties:
+                                type: string
+                              description: Labels used to target Kubernetes nodes
+                              type: object
+                            name:
+                              description: Name of the Rack
```
cscetbon: Please ensure that dc and rack names do not contain a dash as it is used internally as a delimiter. See #158

pkg/apis/db/v1alpha1/cassandracluster_types.go
```diff
@@ -682,9 +683,10 @@ type CassandraClusterSpec struct {
 	MaxPodUnavailable int32 `json:"maxPodUnavailable"` //Number of MaxPodUnavailable used in the PDB
 
 	//Very special Flag to hack CassKop reconcile loop - use with really good Care
-	UnlockNextOperation bool `json:"_unlockNextOperation,omitempty"`
+	UnlockNextOperation bool `json:"unlockNextOperation,omitempty"`
 
 	//Define the Capacity for Persistent Volume Claims in the local storage
+	// +kubebuilder:validation:Pattern=^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
```
cscetbon: What is that comment ?

deploy/crds/db_v1alpha1_cassandracluster_crd.yaml
jsanda: rien qu'un peu :)

deploy/crds/db_v1alpha1_cassandracluster_crd.yaml
jsanda: I added validation regexes

pkg/apis/db/v1alpha1/cassandracluster_types.go
jsanda: Based on my understanding of the code `DataQuantity` ends up being treated as a `Quantity` just like the `CPUAndMem.CPU` and `CPUAndMem.CPU` fields which have the same regex. See https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L30 for more details.

PR #124, opened 2019-10-09
add annotation for pod & services

allamand:
The purpose of this PR is to be able to set specific labels on pods and on the headless service used to access pods.

I needed to be able so specify annotation on the service in order to expose my pods using external-dns.

Examples of annotation needed in the CassandraCluster Spec:
```
  service:
    annotations:
      external-dns.alpha.kubernetes.io/hostname: my.custom.domain.com.

```


pkg/apis/db/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -374,6 +395,29 @@ func (in RackSlice) DeepCopy() RackSlice {
...
+	*out = *in
+	if in.Annotations != nil {
+		in, out := &in.Annotations, &out.Annotations
+		*out = make(map[string]string, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
+		}
+	}
+	return
+}
```
cscetbon: We should not have the same code for both pod and service policies. Let’s refactor it

allamand:
I follow what is done in other operators, and I would like to go further, and move in `spec.pod` section all related stuffs to Pods which are for now under `Spec` only.

But this would be a breaking changes with priori version of operator, but since we are still in alpha I think it's better to do it now. let's discuss this on next meeting.

allamand:
see https://github.com/coreos/etcd-operator/blob/a2c86022d419a73aecedb48e3a404d1a9e4bfba9/doc/user/spec_examples.md#custom-pod-annotations

allamand:
@cscetbon can you approve ?

PR #127, opened 2019-10-11
fix issue where several was updated in all racks at the same time

### LINKED ISSUE #126: Error in operation sequence when updating Cassandra statefulset Envs in CassKop
When Updating Envs in the Statefulset definition, then All statefulsets are updating at the same time.

This is not something wanted as we always want to make only 1 operation at a time.

allamand:
Fix #126


pkg/controller/cassandracluster/generator.go
```diff
@@ -499,10 +499,6 @@ func createEnvVarForCassandraContainer(cc *api.CassandraCluster, status *api.Cas
 			Name:  "CASSANDRA_CLUSTER_NAME",
 			Value: name,
 		},
-		v1.EnvVar{
-			Name:  "CASSANDRA_AUTO_BOOTSTRAP",
-			Value: "true",
-		},
 		v1.EnvVar{
```
cscetbon: 👍

pkg/controller/cassandracluster/cassandra_status.go
```diff
@@ -290,8 +290,9 @@ func UpdateStatusIfSeedListHasChanged(cc *api.CassandraCluster, dcRackName strin
 	}
 
 	//We get a Manual Change on the SeedList
-	//If SeedList has changed in the CRD, we flag the rack with UpdateSeedList Operation To-Do
-	//Once all racks will be enabled with UpdateSeedList=To-do, then we update to ongoing and start the rollUpgrade
+	//If SeedList has changed in the CRD, we flag the rack with UpdateSeedList Operation Configuring
```
cscetbon: Fix case used in sentence. Too many uppercase here and there

pkg/apis/db/v1alpha1/cassandracluster_types.go
```diff
@@ -89,6 +89,9 @@ const (
 	OperationDecommission    string = "decommission"
 	OperationRebuild         string = "rebuild"
 	OperationRemove          string = "remove"
+
+	BreakResyncLoop    bool = true
+	ContinueResyncLoop bool = false
```
cscetbon: Remove types, they are not needed

PR #130, opened 2019-10-17
Better work with cassandra image

allamand:
* add script for liveness probe
    * Remove Env var injected in Cassandra Pod since they are onbly used in Bootstrap Init Container
    * Disable entrypoint for native docker image since it breaks configuration done by Bootstrap Container


docker/bootstrap/files/pre_stop.sh
```diff
@@ -17,18 +17,18 @@
 
 set -e
 
-echo "running PreStop Hook" > /proc/1/fd/1
-echo "**disablegossip**" > /proc/1/fd/1
+echo "running PreStop Hook"
+echo "**disablegossip**"
 nodetool disablegossip
 sleep 1
-echo "**disablebinary**" > /proc/1/fd/1
```
allamand: I remove this because we can't use it in ReadOnluy Container.
But this is no more logs in the output of the container.
@cscetbon If you think at a solution to allow to log that in the output that would be great :)

docker/cassandra/Dockerfile
```diff
@@ -3,3 +3,8 @@ FROM cassandra:3.11
 RUN  apt-get update && apt-get -qq -y install \
        libcap2-bin \
      && setcap cap_ipc_lock=ep $(readlink -f $(which java))
+
+#Make sure we don't use image entrypoint (the bootstrap is done by CassKop)
+ENTRYPOINT []
+#Just start Cassandra
```
allamand: In order to avoid the default image to configure /etc/cassandra when starting (and because I've already done it in the bootstrap phase, I disable it here)
@cscetbon It would be better to disable this in the Operator no ?

docker/cassandra/Dockerfile
```diff
@@ -3,3 +3,8 @@ FROM cassandra:3.11
...
+CMD ["cassandra", "-f"]
```
allamand: also maybe the operator can specify this command by default WDYT ?

docker/cassandra/Makefile
```diff
@@ -14,37 +14,44 @@
 
 PROJECT_ID?=orangeopensource
 ifeq ($(CI_REGISTRY_IMAGE),)
-	PROJECT?=${PROJECT_ID}/cassandra
+	PROJECT?=${PROJECT_ID}/cassandra-image
```
allamand: because it's open on artifactory ;)

documentation/development.md
```diff
@@ -339,6 +347,46 @@ make helm-package
 then add to git the package and make a PR on the repo.
 
 
+## With OLM (Operator Lifecycle Manager)
```
allamand: Work in progress on OLM..

allamand:
rebasing on master

docker/cassandra/Dockerfile
allamand: Yes Tests e2e where not passing because of entrypoint enabled by default in the official wassandra image

docker/cassandra/Dockerfile
allamand: also now I don't add entrypoint and cmd here

PR #131, opened 2019-10-22
[WIP] expose operator and custom resource (CR) metrics

jsanda:


jsanda:
@allamand 

I deployed a Prometheus instance for some manual testing. Here is a screenshot of metrics that get scraped:

prometheus-metrics

It already includes a lot of metrics for things like:

* reconciliation time
* GC
* memory usage
* work queue

You may notice a metric in the screenshot named `goobers_total`. I added that example from [the docs](https://book.kubebuilder.io/reference/metrics.html) to get more familiar with the process for adding custom metrics. It is pretty straightforward. Here is the relevant code:

```go
func init() {
	goobers = prometheus.NewCounter(
		prometheus.CounterOpts{
			Name: "goobers_total",
			Help: "Number of goobers proccessed",
		},
	)
	metrics.Registry.MustRegister(goobers)
}

func (rcc *ReconcileCassandraCluster) Reconcile(request reconcile.Request) (reconcile.Result, error) {
        ...
	goobers.Inc()
       ...
}
```
The `CreateMetricsService` and `CreateServiceMonitors` functions that are called in `main.go` are convenient because they encapsulate all of the work for creating the respective objects. We do trade some flexibility for that convenience though. For example, the label selector is not configurable. It gets set to `name: cassandra-k8s-operator` which may be perfectly fine.

I need to revert my changes for testing in `deploy/operator.yaml`. 

I will update the helm stuff so that the chart sets the metric ports and makes them configurable.

Are there any other changes that are needed for this PR?

allamand:
Hi @jsanda, that's really cool :).

I'm trying to think about which metrics can be useful to manage using this feature.
this is not consolidated, for now just live thinkning :)

maybe we could add metrics for:
- numbers of operation done by Casskop (at cluster and pod level)
- metric that show if there is disruption on the cluster
- basic metrics such as phase, status of the cassandracluster ?

@cscetbon @pierdobauce @fdehay do you see other CassKop metrics that should be useful to see in in prometheus ?

cscetbon:
@allamand @jsanda, maybe : 
- Number of clusters managed by the operator
- Number of errors met
- latency of each loop

Things like that ^

deploy/operator.yaml
```diff
@@ -21,10 +21,13 @@ spec:
       containers:
         - name: cassandra-k8s-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: jsanda/cassandra-k8s-operator:0.4.1-operator-metrics
```
fdehay: We'd rather keep the REPLACE_IMAGE line for the merge. This looks like your own testing image, isn't it?

jsanda:
I am closing this PR as it is out of data and superseded by #198

PR #133, opened 2019-10-23
helm: Adds crds folder in order to install crds with helm v3

### LINKED ISSUE #132: Helm3 CRD Support
Helm v2 has now reached its latest release, and ongoing development is going toward Helm v3.

Some changes happened between v2 and v3, such as there is no more `crd-install` hook, but now CRD should live in its own top-level directory named `crd`, and everything inside it would be installed before rendering templates.

What would be the proper was to support both worlds? Duplicating CRDs (one definition living in `crd` directory and the other one using `crd-install` hook, leading to a warning with helmv3, but should still be working)?

kri5:
There was a breaking change between helm v2 and v3 with crd management.
Helm v2 was using a "crd-install" hook, that no longer exists.
Instead, Helm v3 will install all resources contained in the "crds"
directory before rendering templates.

Fixes #132

### DIFF
--- a/helm/cassandra-operator/Chart.yaml
+++ b/helm/cassandra-operator/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 description: A Helm chart for CassKop - the Orange Cassandra Kubernetes operator
 name: cassandra-operator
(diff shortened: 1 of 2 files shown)


kri5:
This breaks compatibility with Helm 2. Closing as of now.

Issue #123, opened 2019-10-08
Merge CRD fields baseImage & version into cassandraImage 


allamand:
Since we have introduced 2 new parameters for init containers images: ** BootstrapImage** and **InitContainerImage** , I would like to have only 1 parameter for cassandraImage instead of two.

https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/master/pkg/apis/db/v1alpha1/cassandracluster_types.go#L636

But this is a breaking change and will need to rewrite existing CassandraCluster Custom Ressources.

I don't really know what will be the best way to handle the moove.
- keep old parameters and make the operator make the moove to new one ?
- simply remove old and ask user to change CRD definition

WDYT ?

jsanda:
Hi @allamand 

Here are some thoughts:

**Option 1**
The API version is alpha, which is supposed to mean changes can break compatibility. Removing the old fields and making users responsible for handling the change would be the easiest approach. This works for me personally as I am only using the operator for development and testing. If someone is using the operator in production though, this approach may not be desirable.

**Option 2**
Keep the API version the same, replace the fields, and provide a way for users to upgrade their `CassandraCluster` CRs. Maybe this could be done using https://github.com/kubernetes-sigs/kube-storage-version-migrator. It requires the API server to be at v1.15. Another option could be to provide some tooling to do the migration. This could be done with a shell script using `kubectl`. We could alternatively implement a k8s job that does the migration. I personally like the k8s job idea.

**Option 3**
Preserve backwards compatibility. I think that this should result in a new API version, e.g., `v1alpha2` or `v1beta1`. Then we would need a conversion webhook to perform the conversion mappings.

Oh and by the way, happy to help work on this :)

jsanda:
I went ahead and opened https://github.com/Orange-OpenSource/cassandra-k8s-operator/pull/137.

Let me elaborate why I prefer a k8s job vs a shell script for option 2. For a user deploying through Helm, there would be an additional step of running a post-upgrade script. As far as I know, the script could not be executed through Helm. The Helm chart can however be modified to deploy a job that will update any existing `CassandraCluster` CRs. It would be seamless for the end user.

cscetbon:
I vote for **Option 1** as it's still an alpha version. We're also talking about migrating to Helm 3. I don't know how it could help in the future. I think it's a subject to look into for future breaking changes but not in our current situation.

PR #137, opened 2019-10-23
Replace BaseImage and Version fields in spec with CassandraImage

jsanda:
This PR is for #123.

I looked through the e2e tests, and I don't think they require any changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	contrib.go.opencensus.io/exporter/ocagent v0.4.12 // indirect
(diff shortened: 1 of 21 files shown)


jsanda:
I am not yet familiar with the dgoss tests, but they passed for me locally which makes me wonder if the CircleCI failure was a flake. 

@allamand / @cscetbon can you please restart the CI build?

allamand:
@jsanda sometimes dgoss tests failed when trying to reach /metrics enpoint for prometheus. Don't know why, maybe it can takes more time for it to be available in some cases. relaunching the test generally make it pass, but I may need to change some thing maybe in the goss_wait.yaml for this test

PR #138, opened 2019-10-24
controller: Adds support for Tolerations

### LINKED ISSUE #136: Tolerations Support
It would be very interesting to be able to support tolerations in order to schedule cassandra workloads only on some dedicated nodes (in combination with nodeselector).

What would be required to do so?

kri5:
This allows to adds tolerations to the pod template
of the statefulset created by the operator.
It is necessary if you want to allow scheduling thoses pods
on tainted nodes.

Fixes #136

### DIFF
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -250,8 +250,10 @@ func generateCassandraStatefulSet(cc *api.CassandraCluster, status *api.Cassandr
 	rollingPartition := cc.GetRollingPartitionPerRacks(dcRackName)
 	terminationPeriod := int64(api.DefaultTerminationGracePeriodSeconds)
 	var annotations = map[string]string{}
+	var tolerations = []v1.Toleration{}
 	if cc.Spec.Pod != nil {
 		annotations = cc.Spec.Pod.Annotations
+		tolerations = cc.Spec.Pod.Tolerations
 	}
 
 	ss := &appsv1.StatefulSet{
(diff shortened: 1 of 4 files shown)


pkg/controller/cassandracluster/generator.go
```diff
@@ -250,8 +250,10 @@ func generateCassandraStatefulSet(cc *api.CassandraCluster, status *api.Cassandr
 	rollingPartition := cc.GetRollingPartitionPerRacks(dcRackName)
 	terminationPeriod := int64(api.DefaultTerminationGracePeriodSeconds)
 	var annotations = map[string]string{}
+	var tolerations = []v1.Toleration{}
 	if cc.Spec.Pod != nil {
 		annotations = cc.Spec.Pod.Annotations
```
allamand: We should test for spec.pod.toleration and spec.pod.annotation separately

kri5: Agreed, but the test is currently written to test the whole statefulset generation, thus why i added this here. What would you suggest to be tested separatly? What should be duplicated/kept/taken out of the current test?

allamand:
Just wonder what would happened here if there is a pod section with annotation and no toleration for instance ?

kri5:
Indeed. But i just mimicked what was already there for annotations.
There seems to be no test to verify without annotations either.

Should there be more test data in order to test this properly?

If so, is it in the scope of this PR?

allamand:
You’re right ;). I would say that as you touch this part for this PR the problem of code may be improve at the same time. But this could be done in separate PR if you prefere

kri5:
I'd prefer doing this separatly, to keep PR as simple a possible, with the minimal scope.

Issue #143, opened 2019-11-08
Manual update of the seedlist is no more possible


allamand:
The update of the seedlist is correctly mark to configure in each of the racks, but the Flip can't be triggered becasue of a wrong test on it : 

https://github.com/Orange-OpenSource/cassandra-k8s-operator/blob/master/pkg/controller/cassandracluster/reconcile.go#L588
```
	//if global status is not yet  "Configuring", we skip this one
	if cc.Spec.AutoUpdateSeedList &&
		status.LastClusterAction == api.ActionUpdateSeedList &&
		status.LastClusterActionStatus == api.StatusConfiguring {
		var setOperationOngoing = true
```

We need to remove the test on **AutoUpdateSeedList**

PR #146, opened 2019-11-19
Fixes 143 Manual seedlist update

fdehay:
Hello,
SImple fix here with a new unit test. Please check scope of test is enough as I don't use any statefulset
Change consists of removing a test on AutoUpdateSeedlist == true in the function that was blocking the "new" AutoSeedList=False possibility


pkg/controller/cassandracluster/reconcile_test.go
```diff
@@ -165,6 +165,83 @@ func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2(t *testing.T) {
 	assert.Equal(true, reflect.DeepEqual(b, status.SeedList))
 }
 
+func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2ManualSeedList(t *testing.T) {
+	assert := assert.New(t)
+
+	_, cc := helperInitCluster(t, "cassandracluster-2DC-ManualSeedlist.yaml")
```
cscetbon: I think it makes more sense to build the cluster from an existing sample file as we don't want that option to be used beside multicasskop.

pkg/controller/cassandracluster/reconcile_test.go
```diff
@@ -165,6 +165,83 @@ func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2(t *testing.T) {
...
+	cc.Spec.AutoUpdateSeedList = false
+
+	assert.Equal(api.ClusterPhaseInitial, cc.Status.CassandraRackStatus["dc1-rack1"].CassandraLastAction.Name)
+	assert.Equal(api.ClusterPhaseInitial, cc.Status.CassandraRackStatus["dc1-rack2"].CassandraLastAction.Name)
+	assert.Equal(api.ClusterPhaseInitial, cc.Status.CassandraRackStatus["dc2-rack1"].CassandraLastAction.Name)
+	assert.Equal(3, len(cc.Status.CassandraRackStatus))
+
+	//cc.Status.SeedList = cc.InitSeedList()
+
+	var a = []string{
```
cscetbon: Let's use a variable name that has a meaning

pkg/controller/cassandracluster/reconcile_test.go
```diff
@@ -165,6 +165,83 @@ func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2(t *testing.T) {
...
+	UpdateCassandraClusterStatusPhase(cc, status)
+
+	//Flip with AutoUpdateSeedList= true -> update status
+	FlipCassandraClusterUpdateSeedListStatus(cc, status)
+
+	assert.Equal(api.StatusToDo, status.CassandraRackStatus["dc1-rack1"].CassandraLastAction.Status)
+	assert.Equal(api.StatusToDo, status.CassandraRackStatus["dc1-rack2"].CassandraLastAction.Status)
+	assert.Equal(api.StatusToDo, status.CassandraRackStatus["dc2-rack1"].CassandraLastAction.Status)
+
+	//Simulate the Update of SeedList (field CASSANDRA_SEEDLIST of init-container bootstrap
```
cscetbon: Let's get rid of all those comments that are useless

pkg/controller/cassandracluster/testdata/cassandracluster-2DC-ManualSeedlist.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: dc2
+        nodesPerRacks: 1
+        labels: 
+          location.dfy.orange.com/site : mts
+        rack:
+          - name: rack1
+            labels: 
+              location.dfy.orange.com/street : street3
+status:
+  seedlist:
```
cscetbon: Like I said earlier I think we can get rid of that sample deployment file. However I have a question. Why is the seedlist in the status key ? It's surprising ...

cscetbon:
@fdehay, if I understand correctly, the function is already used by other tests. If I'm right, then testing the function itself is enough IMHO. Can you also update the description to add more information about the change ?

pkg/controller/cassandracluster/reconcile_test.go
fdehay: Ok it makes sense and it will prevent double update in case of change that affects these files. Will correct asap

pkg/controller/cassandracluster/reconcile_test.go
fdehay: ok

pkg/controller/cassandracluster/testdata/cassandracluster-2DC-ManualSeedlist.yaml
fdehay: Yes it was the subjezct of the issue 144 which was closed by JAL without answering the initial question.
We discussed it with Seb and thought that to put it in Spec would be too much of a giveway for non casskop control of the seedlist. So we left it in the status key

pkg/controller/cassandracluster/reconcile_test.go
fdehay: I left them to show what the original test was doing after the modified function. Will remove them of course

pkg/controller/cassandracluster/reconcile_test.go
```diff
@@ -177,20 +177,15 @@ func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2ManualSeedList(t *testi
...
-
-	var a = []string{
+	var firstSeedList = []string{
 		"cassandra-demo-dc1-rack1-0.nodomain.com",
 		"cassandra-demo-dc1-rack2-0.nodomain.com",
 		"cassandra-demo-dc2-rack1-0.nodomain.com",
 	}
+	cc.Status.SeedList = firstSeedList
 
 	assert.Equal(3, len(cc.Status.SeedList))
```
cscetbon: @fdehay, not sure this test makes sense here. How could this list change after we've set it ?

pkg/controller/cassandracluster/testdata/cassandracluster-2DC-ManualSeedlist.yaml
Orange-cscetbon: gotcha

pkg/controller/cassandracluster/reconcile_test.go
fdehay: The copy of the initial status is made just before (line 189), so I thought we could simulate an update by changing it just after this copy. We then have a difference between the copier status and the cc.status.seedlist
Thanks for your time here, that's what I was looking for, I am not sure if what I did made sense

fdehay:
Fixes #143

PR #149, opened 2019-11-26
Get rid of env var SERVICE_NAME

cscetbon:


cscetbon:
@allamand I see. I even wonder why we still remove the node from its seeds list. I need to check if it's still needed

PR #153, opened 2019-12-10
Allow Istio to work with Cassandra and encrypt native connections

cscetbon:
This PR removes Intranode and JMX ports from the Service, this way they are not proxyfied by Istio. When Istio proxyfies internodes ports it prevents local connections which is needed by Gossip protocol

```
HOST:PORT                                                           STATUS       SERVER      CLIENT           AUTHN POLICY                                 DESTINATION RULE
cassandra-e2e-exporter-jmx.cassandra-e2e.svc.cluster.local:9500     OK           STRICT      ISTIO_MUTUAL     /default                                     istio-system/default
cassandra-e2e.cassandra-e2e.svc.cluster.local:9042                  OK           STRICT      ISTIO_MUTUAL     /default                                     istio-system/default
cassandra-reaper.cassandra-e2e.svc.cluster.local:8080               OK           STRICT      ISTIO_MUTUAL     /default                                     istio-system/default
```

The proxy configuration that shows the TLS context can be found at https://pastebin.com/raw/jEmxXpwv

### DIFF
--- a/docker/bootstrap/Dockerfile
+++ b/docker/bootstrap/Dockerfile
@@ -48,10 +48,6 @@ LABEL \
     org.label-schema.usage='/README.md'
 
 
-#        libcap2-bin \
-#        procps \
-#        dnsutils && \
-
 RUN mkdir -p $CASSANDRA_DATA  $CASSANDRA_CONF $BOOTSTRAP_CONF $BOOTSTRAP_LIBS $CASSANDRA_LIBS /tmp && \
     set -ex; \
     apt-get update && apt-get -qq -y install --no-install-recommends \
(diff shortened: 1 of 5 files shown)


fdehay:
Question: if we remove the jmx port from the service, will we be able to:
- have C* Reaper connect to the nodes?
- have the operator contact the nodes for jolokia commands?

cscetbon:
> Question: if we remove the jmx port from the service, will we be able to:
> 
> * have C* Reaper connect to the nodes?
> * have the operator contact the nodes for jolokia commands?
Ports on pods are still exposed to the outside world

PR #162, opened 2019-12-26
Adapt CI pipeline for multi-CassKop

erdrix:
Working on circle-ci pipeline, to adapt it to support multi-casskop


.circleci/config.yml
```diff
@@ -294,19 +401,19 @@ workflows:
...
           requires:
-            - unit-test
+            - unit-test-casskop
           filters:
             tags:
               only: /.*/
             branches:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
-              ignore: /pull\/[0-9]+/
+              ignore: /pull\/[0-9w]+/
```
cscetbon: @erdrix, not sure that's what you want here. Are you saying we have branches that have digits and `w`s only ?

.circleci/config.yml
```diff
@@ -366,10 +473,60 @@ workflows:
...
             tags:
               only: /.*/
             branches:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
               ignore: /pull\/[0-9]+/
+
+  # CassKop operator workflow.
+  ci-multicasskop-operator:
+    jobs:
+      - operator/generate:
```
cscetbon: @erdrix, those (generate) should not be jobs but steps

.circleci/config.yml
```diff
@@ -82,207 +84,312 @@ orbs:
...
 
-      - unit-test:
+      - operator/build:
+          name: build-casskop
           requires:
-            - build
+            - unit-test-casskop
           filters:
             tags:
               only: /.*/
```
cscetbon: @erdrix, Our CI config is getting too long and complicated. Use [anchors](https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html) as much as possible to refactor it and avoid repeating filters, steps etc...

multi-casskop/Makefile
```diff
@@ -222,3 +231,8 @@ endif
 .PHONY: release
 release: tag image publish
 
+unit-test-with-vendor:
+	echo TODO
+#	$(UNIT_TEST_CMD_WITH_VENDOR) && echo "success!" || { echo "failure!"; cat test-report.out; exit 1; }
+#	cat test-report.out
+#	$(UNIT_TEST_COVERAGE)
```
cscetbon: Let's get rid of comments

.circleci/config.yml
```diff
@@ -366,10 +473,60 @@ workflows:
...
+      - operator/generate:
+          name: generate-multicasskop
+          operatorDir: cassandra-k8s-operator/multi-casskop
+          operatorName: MultiCassKop
+          requires:
+            - generate-casskop
+          filters:
+            tags:
+              only: /.*/
+#      - operator/unit-test:
```
cscetbon: unused

cscetbon:
@erdrix I know it's a draft but I started to review it just in case it helps

.circleci/config.yml
erdrix: No, it was a mistake !

.circleci/config.yml
erdrix: But something is not clear to me, how can we publish a docker image on the docker hub, before having played the unit-tests ?To me, it doesn't make sense, but maybe I'm missing something ?

.circleci/config.yml
erdrix: Add some anchors to refactor yaml.

.circleci/config.yml
cscetbon: It should be published but only as a temporary image or branch release to be used by all tests. Then only when the PR is merged the release (*-master) should be published

cscetbon: We should also have a job to publish a release when we tag it. Probably need a card for that ?

tools/publish_helm_gcs.sh
```diff
@@ -0,0 +1,103 @@
...
+readonly GCS_BUCKET_INCUBATOR=gs://orange-kubernetes-charts-incubator
+
+main() {
+    mkdir -p tmp
+    setup_helm_client
+    authenticate
+
+#    if ! sync_repo stable "$GCS_BUCKET_STABLE" "$STABLE_REPO_URL"; then
+#        log_error "Not all stable charts could be packaged and synced!"
+#    fi
```
cscetbon: Any reason to still keep this ^ ?

erdrix: To uncomment it, once we will have a stable versions.

PR #164, opened 2020-01-09
Rename repository

cscetbon:


erdrix:
You change each `cassandra-k8s-operator` into `casskop`, but for instance the docker repository `orangeopensource/casskop` doesn't exist, it won't work.
So, either we only replace `Orange-OpenSource/cassandra-k8s-operator`, or we need to make sure that every reference to `cassandra-k8s-operator` works well once replaced.

allamand:
It’s easy to create the docker repo Casskop

fdehay:
J'ai pas les droits pour le repo docker il faut un owner de l'organisation orangeopensource, j'ai demandé à lal iste gihub...

Issue #168, opened 2020-01-15
[Multi Casskop] Rebuild operation can't work


erdrix:
I created on GCP two GKE cluster, and deployed on each, a casskop operator and external dns.
On top of that, I deploy the multi-casskop operator, and the following  `Multi-Casskop` resource : 

```yaml
apiVersion: db.orange.com/v1alpha1
kind: MultiCasskop
metadata:
  name: multi-casskop-demo
spec:
  deleteCassandraCluster: true
  base: #<-- Specify the base of our CassandraCluster
    apiVersion: "db.orange.com/v1alpha1"
    kind: "CassandraCluster"
    metadata:
      name: cassandra-demo
      namespace: cassandra-demo
      labels:
        cluster: casskop
    spec:
      cassandraImage: orangeopensource/cassandra-image:3.11
      bootstrapImage: orangeopensource/cassandra-bootstrap:0.1.3
      configMapName: cassandra-configmap-v1
      service:
        annotations:
          external-dns.alpha.kubernetes.io/hostname: casskop.external-dns-test.gcp.trycatchlearn.fr.
      rollingPartition: 0
      dataCapacity: "20Gi"
      dataStorageClass: "standard-wait"
      imagepullpolicy: IfNotPresent
#      imagepullpolicy: Always
      hardAntiAffinity: false
      deletePVC: true
      autoPilot: false
      gcStdout: false
      autoUpdateSeedList: false
      debug: false
      maxPodUnavailable: 1
      nodesPerRacks: 1
      runAsUser: 999
      resources:
        requests: &requests
          cpu: '1'
          memory: 2Gi
        limits: *requests
    status:
      seedlist:   #<-- at this time the seedlist must be fullfilled manually with known predictive name of pods
        - cassandra-demo-dc1-rack1-0.casskop.external-dns-test.gcp.trycatchlearn.fr
        - cassandra-demo-dc1-rack1-1.casskop.external-dns-test.gcp.trycatchlearn.fr
        - cassandra-demo-dc2-rack1-0.casskop.external-dns-test.gcp.trycatchlearn.fr
        - cassandra-demo-dc2-rack1-1.casskop.external-dns-test.gcp.trycatchlearn.fr    

  override: #<-- Specify overrides of the CassandraCluster depending on the target kubernetes cluster
    gke-master-west1-b:
      spec:
        topology:
          dc:
            - name: dc1
              nodesPerRacks: 2
              numTokens: 256
              labels:
[...]

PR #179, opened 2020-01-29
remove check dc from CC object (for mutli-cluster)

erdrix:
Fix issue : https://github.com/Orange-OpenSource/casskop/issues/168, Rebuild from dc operation can't work with multi-casskop setup.


pkg/controller/cassandracluster/pod_operation.go
```diff
@@ -685,9 +685,10 @@ func (rcc *ReconcileCassandraCluster) runRebuild(hostName string, cc *api.Cassan
 
 	if labelSet != true {
 		err = errors.New("operation-argument is needed to get the datacenter name to rebuild from")
-	} else if cc.IsValidDC(rebuildFrom) == false {
-		err = fmt.Errorf("%s is not an existing datacenter", rebuildFrom)
 	}
+	/*} else if cc.IsValidDC(rebuildFrom) == false {
+		err = fmt.Errorf("%s is not an existing datacenter", rebuildFrom)
+	}*/
```
cscetbon: @erdrix As the current CassandraCluster does not know anything about datacenter hosted by another remote object, you propose to remove the check. Shouldn't we test it using JMX instead ? Did you try it with a non existing DC and check logs ? If they are explicit I'm 👍, if not we should test it differently

erdrix: It's not really explicit : 

```logs
$ Cannot rebuild from dc5: java.lang.IllegalStateException : Unable to find sufficient sources for streaming range (3396609007184563874,3447869076743942192] in keyspace system_distributed
```

pkg/controller/cassandracluster/pod_operation.go
```diff
@@ -677,18 +677,21 @@ func (rcc *ReconcileCassandraCluster) runUpgradeSSTables(hostName string, cc *ap
...
 	logrus.WithFields(logrus.Fields{"cluster": cc.Name, "rack": dcRackName, "pod": pod.Name,
 		"hostName": hostName, "operation": operation}).Info("Operation start")
 
+	jolokiaClient, err := NewJolokiaClient(hostName, JolokiaPort, rcc,
+		cc.Spec.ImageJolokiaSecret, cc.Namespace)
+
 	if labelSet != true {
 		err = errors.New("operation-argument is needed to get the datacenter name to rebuild from")
+	} else if keyspaces, err = jolokiaClient.HasDataInDC(rebuildFrom); err == nil  && len(keyspaces) == 0 {
+		err = fmt.Errorf("%s is not has no nonLocalKeyspace defined", rebuildFrom)
```
cscetbon: 2 things : 
- `HasDataInDC` along with its comments should be renamed. It returns a list of keyspaces found at the new current Jolokia server that have data replicated to the provided DC
- `"%s is not has no nonLocalKeyspace defined"` -> `"%s has no keyspace to replicate data from`

Makefile
```diff
@@ -410,7 +410,7 @@ configure-psp:
 # Usage example:
 # REPLICATION_FACTOR=3 make cassandra-stress small
 #
-ifeq (cassandra-stress,$(firstword $(MAKECMDGOALS)))
+ifeq (stress-cassandra,$(firstword $(MAKECMDGOALS)))
```
cscetbon: Ugh, the action here references the tool that is used underneath, which means `cassandra-stress` ...

pkg/controller/cassandracluster/pod_operation.go
```diff
@@ -692,7 +692,7 @@ func (rcc *ReconcileCassandraCluster) runRebuild(hostName string, cc *api.Cassan
 	} else if keyspaces, err = jolokiaClient.HasDataInDC(rebuildFrom); err == nil  && len(keyspaces) == 0 {
 		err = fmt.Errorf("%s is not has no nonLocalKeyspace defined", rebuildFrom)
 	}
-
+
```
cscetbon: no need for those extra spaces

Makefile
erdrix: I had change action name to `stress-cassadra` because of folder name which is `cassandra-stress`. 
If you run `make cassandra-stress` without this change, it makes the operation with folder and not with makefile action

PR #184, opened 2020-02-02
Use Jolokia instead of nodetool in probes

cscetbon:
This PR :
- adds curl in bootstrap image (new mount point tools in order to host it)
- rename file ready-probe.sh to readiness-probe.sh
- fix an issue during the configuration of seeds for the first node
- uses Jolokia queries instead of nodetool calls which are expensive and slow

We'll need to deploy and use a new bootstrap image but I'll wait for the first feedback before pushing it


cscetbon:
@allamand we shouldn’t, I thought that was something that was done in the master branch and was going to ask Alex. But it seems it’s in my branch for some reason. Going to remove it. Good catch !

cscetbon:
> @allamand we shouldn’t, I thought that was something that was done in the master branch and was going to ask Alex. But it seems it’s in my branch for some reason. Going to remove it. Good catch !

fdehay:
While testing a little, I upgraded a cluster with a bootstrap image in 0.1.2 to 0.1.4.
The rack would stay stuck while trying to reboot with a warning saying that:
/etc/cassandra/ready-probe.sh does not exist
Maybe I should not have tested this at this stage but I wanted to share anyway

erdrix:
Did we test the image upgrade from 0.1.3 to 0.1.4 on deployed cluster ? If yes is it seamless ? Maybe we would need to add documentation about migration to the new version, referring to the comment of @fdehay.

cscetbon:
> Did we test the image upgrade from 0.1.3 to 0.1.4 on deployed cluster ? If yes is it seamless ? Maybe we would need to add documentation about migration to the new version, referring to the comment of @fdehay.

@erdrix The only thing is that I renamed one of the probe file for consistency and for this reason there would be a problem if only the bootstrap image is updated, or only casskop. Do we think that we really need to test it knowing that we're still in alpha ? We can add a note on that but not sure it's important.

Readme.md
```diff
@@ -307,6 +307,52 @@ helm install --name casskop casskop/cassandra-operator --set debug.enabled=true
...
+        ....
+        name: cassandra
+        ...
+        volumeMounts:
+        - mountPath: /opt/bin
+          name: tools
+        ....
+  ```
+
+- Do exactly all the same changes in the annotation banzaicloud.com/last-applied
```
cscetbon: @allamand @erdrix can you confirm it's needed ? It was part of all the changes I made in my numerous tests, but if it's not needed let me know.

erdrix: It works for me, without updating banzaicloud part !

Issue #192, opened 2020-02-14
Can't update sidecar list with less items


erdrix:
# Bug Report

## What did you do

I create a CassandraCluster with the following `sidecarConfigs` (3 elements) : 

```yaml 
...
sidecarConfigs:
    - args: ["tail", "-F", "/var/log/cassandra/system.log"]
      image: registry.gitlab.si.francetelecom.fr/dfyarchicloud/dfyarchicloud-registry/alpine-tini
      imagePullPolicy: Always
      name: cassandra-log
      resources:
        limits:
          cpu: 50m
          memory: 50Mi
        requests:
          cpu: 10m
          memory: 10Mi
      volumeMounts:
        - mountPath: /var/log/cassandra
          name: cassandra-logs
    - args: ["tail", "-F", "/var/log/cassandra/gc.log.0.current"]
      image: registry.gitlab.si.francetelecom.fr/dfyarchicloud/dfyarchicloud-registry/alpine-tini
      imagePullPolicy: Always
      name: gc-log
      resources:
        limits:
          cpu: 50m
          memory: 50Mi
        requests:
          cpu: 10m
          memory: 10Mi
      volumeMounts:
        - mountPath: /var/log/cassandra
          name: gc-logs
    - args: ["tail", "-F", "/var/log/cassandra/gc.log.0.current"]
      image: registry.gitlab.si.francetelecom.fr/dfyarchicloud/dfyarchicloud-registry/alpine-tini
      imagePullPolicy: Always
      name: gc-log2
      resources:
        limits:
          cpu: 50m
          memory: 50Mi
        requests:
          cpu: 10m
          memory: 10Mi
...
```
And after I applied an update of `sidecarConfigs` with a lower number of elements : 

```yaml
...
  sidecarConfigs:
    - args: ["tail", "-F", "/var/log/cassandra/system.log"]
      image: registry.gitlab.si.francetelecom.fr/dfyarchicloud/dfyarchicloud-registry/alpine-tini
      imagePullPolicy: Always
      name: cassandra-log
      resources:
        limits:
          cpu: 50m
          memory: 50Mi
        requests:
          cpu: 10m
          memory: 10Mi
      volumeMounts:
        - mountPath: /var/log/cassandra
          name: cassandra-logs
    - args: ["tail", "-F", "/var/log/cassandra/gc.log.0.current"]
      image: registry.gitlab.si.francetelecom.fr/dfyarchicloud/dfyarchicloud-registry/alpine-tini
[...]

PR #193, opened 2020-02-14
Add check on container length for statefulset comparison

erdrix:
Fix issue : https://github.com/Orange-OpenSource/casskop/issues/192

### DIFF
--- a/pkg/controller/cassandracluster/statefulset.go
+++ b/pkg/controller/cassandracluster/statefulset.go
@@ -136,32 +136,37 @@ func statefulSetsAreEqual(sts1, sts2 *appsv1.StatefulSet) bool {
 			return false
 		}
 	*/
-	containersSpecSts1Len := len(sts1.Spec.Template.Spec.Containers)
-	containersSpecSts2Len := len(sts2.Spec.Template.Spec.Containers)
+	sts1Spec := &sts1.Spec.Template.Spec
+	sts2Spec := &sts2.Spec.Template.Spec
+
+	sts1Spec.SchedulerName = sts2Spec.SchedulerName
+	sts1Spec.DNSPolicy = sts2Spec.DNSPolicy // ClusterFirst
+
+	containersSpecSts1Len := len(sts1Spec.Containers)
+	containersSpecSts2Len := len(sts2Spec.Containers)
 	if containersSpecSts1Len != containersSpecSts2Len{
 		logrus.WithFields(logrus.Fields{"statefulset": sts1.Name,
-			"namespace": sts1.Namespace}).Info(fmt.Sprintf("Template is different, the number of containers are not the same: len(sts1.Spec.Template.Spec.Containers) = %d, len(sts2.Spec.Template.Spec.Containers) = %d", containersSpecSts1Len,containersSpecSts2Len))
+			"namespace": sts1.Namespace}).Info(
+			fmt.Sprintf("Template is different, the number of containers are not the same: len(sts1.Spec.Template.Spec.Containers) = %d, " +
+				"len(sts2.Spec.Template.Spec.Containers) = %d",
+				containersSpecSts1Len,containersSpecSts2Len))
 		return false
 	}
 
-	sts1.Spec.Template.Spec.SchedulerName = sts2.Spec.Template.Spec.SchedulerName
-	sts1.Spec.Template.Spec.DNSPolicy = sts2.Spec.Template.Spec.DNSPolicy // ClusterFirst
+	for i := 0; i < len(sts1Spec.Containers); i++ {
+		sts1Spec.Containers[i].LivenessProbe = sts2Spec.Containers[i].LivenessProbe
+		sts1Spec.Containers[i].ReadinessProbe = sts2Spec.Containers[i].ReadinessProbe
 
-
-	for i := 0; i < len(sts1.Spec.Template.Spec.Containers); i++ {
-		sts1.Spec.Template.Spec.Containers[i].LivenessProbe = sts2.Spec.Template.Spec.Containers[i].LivenessProbe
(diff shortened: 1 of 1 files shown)


pkg/controller/cassandracluster/statefulset.go
```diff
@@ -136,8 +136,18 @@ func statefulSetsAreEqual(sts1, sts2 *appsv1.StatefulSet) bool {
...
+	containersSpecSts2Len := len(sts2.Spec.Template.Spec.Containers)
+	if containersSpecSts1Len != containersSpecSts2Len{
+		logrus.WithFields(logrus.Fields{"statefulset": sts1.Name,
+			"namespace": sts1.Namespace}).Info(fmt.Sprintf("Template is different, the number of containers are not the same: len(sts1.Spec.Template.Spec.Containers) = %d, len(sts2.Spec.Template.Spec.Containers) = %d", containersSpecSts1Len,containersSpecSts2Len))
+		return false
+	}
+
 	sts1.Spec.Template.Spec.SchedulerName = sts2.Spec.Template.Spec.SchedulerName
 	sts1.Spec.Template.Spec.DNSPolicy = sts2.Spec.Template.Spec.DNSPolicy // ClusterFirst
+
```
cscetbon: Let's remove that extra line.

pkg/controller/cassandracluster/statefulset.go
```diff
@@ -136,8 +136,18 @@ func statefulSetsAreEqual(sts1, sts2 *appsv1.StatefulSet) bool {
 			return false
 		}
 	*/
+	containersSpecSts1Len := len(sts1.Spec.Template.Spec.Containers)
+	containersSpecSts2Len := len(sts2.Spec.Template.Spec.Containers)
```
cscetbon: I propose you create 2 new variables to store `sts1.Spec.Template.Spec.Containers[I]` and `sts2.Spec.Template.Spec.Containers`. Cause it's also used later to access each container in those lists. That would make the code more readable.

Why not just use what we proposed in the past using a diff library ? I use `github.com/r3labs/diff` in a unit test but there might be a better approach.

pkg/controller/cassandracluster/statefulset.go
```diff
@@ -136,8 +136,18 @@ func statefulSetsAreEqual(sts1, sts2 *appsv1.StatefulSet) bool {
...
+	if containersSpecSts1Len != containersSpecSts2Len{
+		logrus.WithFields(logrus.Fields{"statefulset": sts1.Name,
+			"namespace": sts1.Namespace}).Info(fmt.Sprintf("Template is different, the number of containers are not the same: len(sts1.Spec.Template.Spec.Containers) = %d, len(sts2.Spec.Template.Spec.Containers) = %d", containersSpecSts1Len,containersSpecSts2Len))
```
cscetbon: This line is too long and should be split

pkg/controller/cassandracluster/statefulset.go
erdrix: I tried to use the `patch` module from **banzaicloud** : 

```go
	desiredType := reflect.TypeOf(sts2)
	patchResult, err := patch.DefaultPatchMaker.Calculate(sts1, sts2)

	if err != nil {
		log.Error(err, "could not match objects", "kind", desiredType)
	} else if !patchResult.IsEmpty(){
		logrus.WithFields(logrus.Fields{"statefulset": sts1.Name,
			"namespace": sts1.Namespace}).Info("resource diffs",
			"patch", string(patchResult.Patch),
			"current", string(patchResult.Current),
			"modified", string(patchResult.Modified),
			"original", string(patchResult.Original))
		return false
	}
```

But he find systematically a difference in the `volumeClaimTemplate` : 

```json 
{
    "spec": {
        "volumeClaimTemplates": [
            {
                "metadata": {
                    "labels": {
                        "app": "cassandracluster",
                        "cassandracluster": "cassandra-bgl",
                        "cassandraclusters.db.orange.com.dc": "dc3",
                        "cassandraclusters.db.orange.com.rack": "rack1",
                        "cluster": "casskop",
                        "dc-rack": "dc3-rack1"
                    },
                    "name": "data"
                },
                "spec": {
                    "accessModes": [
                        "ReadWriteOnce"
                    ],
                    "resources": {
                        "requests": {
                            "storage": "20Gi"
                        }
                    },
                    "storageClassName": "local-storage"
                },
                "status": {}
            },
            {
                "metadata": {
                    "labels": {
                        "app": "cassandracluster",
                        "cassandracluster": "cassandra-bgl",
                        "cassandraclusters.db.orange.com.dc": "dc3",
                        "cassandraclusters.db.orange.com.rack": "rack1",
                        "cluster": "casskop",
                        "dc-rack": "dc3-rack1"
                    },
                    "name": "gc-logs"
                },
                "spec": {
                    "accessModes": [
                        "ReadWriteOnce"
                    ],
                    "resources": {
                        "requests": {
                            "storage": "10Gi"
                        }
                    },
                    "storageClassName": "local-storage"
                },
                "status": {}
            },
            {
                "metadata": {
                    "labels": {
                        "app": "cassandracluster",
                        "cassandracluster": "cassandra-bgl",
                        "cassandraclusters.db.orange.com.dc": "dc3",
                        "cassandraclusters.db.orange.com.rack": "rack1",
                        "cluster": "casskop",
                        "dc-rack": "dc3-rack1"
                    },
                    "name": "cassandra-logs"
                },
                "spec": {
                    "accessModes": [
                        "ReadWriteOnce"
                    ],
                    "resources": {
                        "requests": {
                            "storage": "10Gi"
                        }
                    },
                    "storageClassName": "local-storage"
                },
                "status": {}
            }
        ]
    }
}
```

So I think that it doesn't work well for now :/

PR #198, opened 2020-02-25
Add custom metrics to operator

cscetbon:


pkg/controller/cassandracluster/reconcile.go
```diff
@@ -625,6 +670,9 @@ func FlipCassandraClusterUpdateSeedListStatus(cc *api.CassandraCluster, status *
 					logrus.WithFields(logrus.Fields{"cluster": cc.Name,
 						"dc-rack": dcRackName}).Infof("Update Rack Status UpdateSeedList=ToDo")
 					dcRackStatus.CassandraLastAction.Name = api.ActionUpdateSeedList
+					ClusterAction.With(
+						prometheus.Labels{"cluster": cc.Name},
+					).Set(api.ClusterActionVal(api.ActionUpdateSeedList))
```
erdrix: I wonder if this call could not be a more simple function such as : 

```go
func (clusterAction *GaugeVec) updateClusterAction(clusterName string, action Action) {
  clusterAction.With(prometheus.Labels{"cluster": clusterName},)
    .Set(api.ClusterActionVal(string(action)))
}
```

To make the code more readable.

In the same way, I think we need to change the Action, ClusterPhase, Status and Operation type, to something more restrictive than a String. Just as they did it in this operator : https://github.com/banzaicloud/kafka-operator/blob/master/api/v1beta1/common_types.go#L76

It makes the code more readable, and usage simplier.

pkg/apis/db/v1alpha1/cassandracluster_types.go
```diff
@@ -93,6 +94,31 @@ const (
...
+			return i
+		}
+	}
+	return -1
+}
+
+// ClusterPhaseVal returns corresponding float64 value
+func ClusterPhaseVal(clusterPhase string) float64 {
+	return float64(
+		index([]string{ClusterPhaseInitial, ClusterPhaseRunning, ClusterPhasePending}, clusterPhase),
```
erdrix: As discussed, I think that this approach could lead to a lot of mistakes, due to implicit link between the `ClusterPhase` and the `float` returned. 
It could be have at least two other solutions : 

1. To expose the link using a struct, which will avoid mistakes, when we will introduce new ClusterPhase : 

```go
type ClusterPhase struct {
  Id float `json:"clusterId"`
  Name string `json:name`
}
```

2. To define one metric per ClusterPhase, using a boolean to specify if the cluster is active or not in this phase.

pkg/apis/db/v1alpha1/cassandracluster_types.go
```diff
@@ -134,7 +160,7 @@ func (cc *CassandraCluster) SetDefaults() bool {
 		changed = true
 	}
 	if len(cc.Status.Phase) == 0 {
-		cc.Status.Phase = ClusterPhaseInitial
+		cc.Status.Phase = string(ClusterPhaseInitial)
```
erdrix: Cast string to string ? Reliqua of some test ?

cmd/manager/main.go
```diff
@@ -181,6 +183,8 @@ func main() {
 
 	logrus.Info("Registering Components.")
 
+	prometheusMetrics.Registry.MustRegister(cassandracluster.ClusterPhase, cassandracluster.ClusterAction)
```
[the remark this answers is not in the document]
cscetbon: Yeah you missed it but it's similar to what I'm doing. However I'm not using a second port to expose the metrics I'm adding. That's why I don't have the function `serveCRMetrics`.

pkg/controller/cassandracluster/reconcile.go
cscetbon: Okay for the first part. For the second part I don't know how difficult it would be as it ends in the status of the object. I need to take a closer look to the kafka operator I guess. Let's chat about it tomorrow

pkg/apis/db/v1alpha1/cassandracluster_types.go
cscetbon: Yes I think we agree. I wanted to avoid having to duplicate code but having a place in the code where we can see the actual value definitely makes sense.
I don't understand your second point.

pkg/apis/db/v1alpha1/cassandracluster_types.go
cscetbon: yeah probably. Or I didn't push some changes

pkg/apis/db/v1alpha1/cassandracluster_types.go
cscetbon: @erdrix why do you use marshable syntax in the type definition ? Because of the second point ?

erdrix: The second point was to have for each ClusterAction / ClusterPhase type a boolean metric. 
For the marshable syntax, I make a quick copy/paste and forgot to remove it ^^

PR #200, opened 2020-02-26
catch nil pvcSpec error

erdrix:
Catch StorageConfig error, when generating PVC from a nil PvcSpec (which cause nil pointer exception).


cscetbon:
@erdrix Can you add at least a tests to check those 2 new cases ?

PR #201, opened 2020-03-06
add liveness and readiness probe configurable in CassandraCluster object

erdrix:
Add the ability to configure, through `CassandraCluster` definition, for ReadinessProbe and LivenessProbe of main cassandra conrtainer the following Probe fields : 
- InitialDelaySeconds
- TimeoutSeconds
- PeriodSeconds
- SuccessThreshold
- FailureThreshold


pkg/controller/cassandracluster/testdata/cassandracluster-2DC-default.yaml
```diff
@@ -0,0 +1,98 @@
...
+    - mountPath: "/var/log/cassandra"
+      name: "cassandra-logs"
+      pvcSpec:
+        accessModes:
+          - ReadWriteOnce
+        storageClassName: standard-wait
+        resources:
+          requests:
+            storage: 10Gi
+  sidecarConfigs:
```
cscetbon: @erdrix, we already have https://github.com/Orange-OpenSource/casskop/blob/4a9b6210d4a75086e033c03f32aef7b553715b74/pkg/controller/cassandracluster/testdata/cassandracluster-2DC.yaml that has sidecars. Why that new file ?

erdrix: Because I want a CassandraCluster without liveness and readiness configurated to check in unit tests that the default values are well set in the statefulset.

PR #206, opened 2020-03-18
Fix readiness & liveness probe configuration detection

erdrix:
Reworks statefulset's changes detection by the operator, to work with live & ready probe configurations.


cscetbon:
@erdrix I think it's missing a unit test. See https://github.com/Orange-OpenSource/casskop/blob/f251372e17fa337110ec3b7b82fe56ff1140dba3/pkg/controller/cassandracluster/cassandra_status_test.go#L261 on how to use reconcile if you want to test the update of the new parameters. I let you decide what is the easiest, unit test or e2e

erdrix:
I'm ok with you about unit test, I added some test but not using the reconcile, the bug was about the `StatefulSetsAreEqual` function which didn't detecting update, so I added some unit-tests for this function :) 
Tell me if its seems enough for you.

pkg/controller/cassandracluster/statefulset_test.go
```diff
@@ -0,0 +1,76 @@
...
+	rackName := "rack1"
+	dcRackName := fmt.Sprintf("%s-%s", dcName, rackName)
+
+	_, cc := helperInitCluster(t, "cassandracluster-2DC.yaml")
+	cc.CheckDefaults()
+	labels, nodeSelector := k8s.GetDCRackLabelsAndNodeSelectorForStatefulSet(cc, 0, 0)
+	sts, _ := generateCassandraStatefulSet(cc, &cc.Status, dcName, dcRackName, labels, nodeSelector, nil)
+
+	var testSetup map[string]liveAndReadinessProbeExpected
+	testSetup = make(map[string]liveAndReadinessProbeExpected)
```
cscetbon: just do `testSetup := make(map[string]liveAndReadinessProbeExpected)`

PR #208, opened 2020-04-05
Make operator less verbose

sheerun:
This PR removes info log that constitutes half of my logs and doesn't tell much (`We will list available pod`). I don't think logging get requests is necessary, only errors or commands.

What do you think?


sheerun:
Autoformatter also changed few other lines in this file

sheerun:
Another option: make it debug instead of info

cscetbon:
@sheerun Can you make it a debug log instead ? Thanks

PR #224, opened 2020-05-29
Monitoring

cscetbon:
### What's in this PR?
Adds documentation on how to add monitoring and a dashboard to monitor cassandra metrics.

### DIFF
--- a/documentation/monitoring.md
+++ b/documentation/monitoring.md
@@ -3,7 +3,7 @@
 ## Install Prometheus operator
 
 ```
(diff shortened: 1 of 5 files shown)


cscetbon:
@ahmedjami Can you take a look at the dashboard. I think some metrics need to be aggregated. For example the metric **Total space used**, with **only one node**, I get many different values and it doesn't make sense to me:

Screen Shot 2020-05-29 at 7 27 45 PM

cscetbon:
I quickly looked into it and yeah some metrics need to be aggregated before being displayed. For instance, **Total Space Used** queries _cassandra_table_live_disk_space_bytes_ which is a metric **per table** so we get many of them per each node.

cscetbon:
@ahmedjami I made a few more changes to aggregate data. I think data should be still aggregated in other graphs and I don't think it makes sense to have all system tables on them. I can't imagine how each of those graphs would be with 20 nodes.

PR #225, opened 2020-05-31
 Use k3d instead of Minikube

cscetbon:
Let's replace Minikube which takes too long to setup and interact with

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -508,3 +508,8 @@ workflows:
           requires:
             - report-multicasskop
           <<: *only_on_release_tag
+      - deploy-website:
+          name: deploy-website
+          requires:
+             - report-casskop
+          <<: *website_publish
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,29 @@ orbs:
                 mkdir -p ${HOME}/.kube
                 cp $(k3d get-kubeconfig) ${HOME}/.kube/config
           - run:
+              name: Create file with build image version
+              command: |
+                make build-image > casskop-build-image
+          - restore_cache:
+              keys:
+                - '{{ checksum "casskop-build-image" }}'
+          - run:
+              name: Load Casskop build image into k3d
+              command: |
+                if [ -f "casskop-build-image.tgz" ]
+                then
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -445,67 +423,62 @@ workflows:
...
-            - unit-test-casskop
+            - build-casskop
           <<: *ignore_fork_pr_filter
+
       - operator/publish-helm:
           name: publish-helm-orange-incubator-casskop
           requires:
             - report-casskop
           <<: *only_on_release_tag
-      - deploy-website:
```
erdrix, line 501: Why do you remove the website publication ?

.circleci/config.yml
```diff
@@ -9,80 +9,76 @@ experimental:
...
-                  function is_ready()
-                  {
-                    kubectl get nodes -o json \
-                      | jq '.items[].status.conditions[] | select(.type=="Ready" and .status=="True")'
-                  }
-                  until is_ready
-                  do
-                    sleep 1
-                  done
-                EOT
```
erdrix, line 85: No need to wait the nodes for k3d ?

.circleci/config.yml
```diff
@@ -9,80 +9,76 @@ experimental:
...
+              name: Load Casskop build image into k3d
+              command: |
+                if [ -f "casskop-build-image.tgz" ]
+                then
+                  docker load < casskop-build-image.tgz
+                else
+                  docker pull $(cat casskop-build-image)
+                  docker save $(cat casskop-build-image)|gzip > casskop-build-image.tgz
+                fi
+                k3d import-images $(cat casskop-build-image)
```
erdrix, line 66: why do you need to load this image in minikube, and we did'nt have to on minikube ? What is the difference ?

cscetbon: I cache the image instead of pulling it ever time a test start. It should be faster than pulling 1GB I suppose

.circleci/config.yml
cscetbon: We do with —wait 0

.circleci/config.yml
cscetbon: I’ll fix that one

PR #230, opened 2020-06-13
Set bootstrap env vars based on Cassandra resources

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
This PR fixes a bug where the env CASSANDRA_MAX_HEAP is set based on that container resources instead of the main Cassandra container's resources.

### DIFF
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -401,9 +401,20 @@ func generateCassandraStatefulSet(cc *api.CassandraCluster, status *api.Cassandr
 	}
 
 	// Merge cassandra main container environment variables into sidecars.
+
+	var sidecarEnv []v1.EnvVar
+
+	// Collect all env vars from bootstrap container except the heap size
+	for _, env := range bootstrapContainer.Env {
+		if env.Name != cassandraMaxHeap {
+			sidecarEnv = append(sidecarEnv, env)
+		}
+	}
+
+	// Add all those env vars to each sidecar
 	for idx, container := range ss.Spec.Template.Spec.Containers {
 		if container.Name != cassandraContainerName {
-			ss.Spec.Template.Spec.Containers[idx].Env = append(container.Env, bootstrapContainer.Env...)
+			ss.Spec.Template.Spec.Containers[idx].Env = append(container.Env, sidecarEnv...)
 		}
 	}
 
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -45,6 +45,7 @@ type JvmMemory struct {
 const (
 	cassandraContainerName = "cassandra"
 	bootstrapContainerName = "bootstrap"
+	cassandraMaxHeap       = "CASSANDRA_MAX_HEAP"
 	defaultJvmMaxHeap      = "2048M"
 	hostnameTopologyKey    = "kubernetes.io/hostname"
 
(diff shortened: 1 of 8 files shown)


pkg/controller/cassandracluster/generator.go
```diff
@@ -395,7 +401,7 @@ func generateCassandraStatefulSet(cc *api.CassandraCluster, status *api.Cassandr
 	// Merge cassandra main container environment variables into sidecars.
 	for idx, container := range ss.Spec.Template.Spec.Containers {
 		if container.Name != cassandraContainerName {
-			ss.Spec.Template.Spec.Containers[idx].Env = append(ss.Spec.Template.Spec.Containers[idx].Env, cassandraContainer.Env...)
+			ss.Spec.Template.Spec.Containers[idx].Env = append(container.Env, bootstrapContainer.Env...)
```
cscetbon, line 404: @erdrix , do you think we really need it ? We could probably remove some of them and maybe push only DC, RACK and CLUSTER_NAME ? why would they need the heap size, the seeds etc.. ?

erdrix: When I did it, I took the side of pushing them all because we had a low visibility on sidecar usage and therefore the env vars required for these ... So if the necessary are the ones you specified no problems for me :)

pkg/controller/cassandracluster/generator.go
```diff
@@ -577,33 +585,15 @@ func createEnvVarForCassandraContainer(cc *api.CassandraCluster, status *api.Cas
 				},
 			},
 		},
-		v1.EnvVar{
-			Name: "POD_NAME",
```
erdrix, line 581: Did you remove this one because it's unused ?

pkg/controller/cassandracluster/generator.go
```diff
@@ -577,33 +585,15 @@ func createEnvVarForCassandraContainer(cc *api.CassandraCluster, status *api.Cas
...
 			Name:  "CASSANDRA_GC_STDOUT",
 			Value: strconv.FormatBool(cc.Spec.GCStdout),
 		},
-		v1.EnvVar{
+		{
 			Name:  "CASSANDRA_NUM_TOKENS",
 			Value: strconv.Itoa(int(numTokensPerRacks)),
 		},
-		v1.EnvVar{
-			Name: "NODE_NAME",
```
erdrix, line 598: Did you remove this one because it's unused ?

pkg/controller/cassandracluster/generator.go
```diff
@@ -750,7 +739,17 @@ func createCassandraContainer(cc *api.CassandraCluster, status *api.CassandraClu
 				},
 			},
 		},
-		Env: createEnvVarForCassandraContainer(cc, status, resources, dcRackName),
```
erdrix, line 753: There is no need of all the env vars that were push here before ? (Any ideas why they were pushed before ? Just to be sure that we don't miss anything else like I did for the MAX_HEAP :D )

test/e2e/cassandracluster_test.go
```diff
@@ -405,8 +405,8 @@ func cassandraClusterCleanupTest(t *testing.T, f *framework.Framework, ctx *fram
 			return false, nil
 		}
 
-		if !endTime.After(startTime) {
-			t.Logf("Expected endTime (%s) to be after startTime (%s)", endTime, startTime)
+		if endTime.Sub(startTime) < 0 {
```
erdrix, line 408: Why did you change this test ?

pkg/controller/cassandracluster/generator.go
cscetbon: Correct

pkg/controller/cassandracluster/generator.go
cscetbon: yes

pkg/controller/cassandracluster/generator.go
cscetbon: the Cassandra container does not need them

test/e2e/cassandracluster_test.go
cscetbon: because some time it runs even in the same second so startTime = endTime and the test fails.

PR #235, opened 2020-07-10
Fix multi decommissioning 

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

### Why?

When decommissioning multiple nodes, only the first one or at least not all are decommissioning but just removed.

This PR also adds unit tests related to the decommission to ensure it doesn't happen with new changes

### DIFF
--- a/pkg/controller/cassandracluster/decommission_test.go
+++ b/pkg/controller/cassandracluster/decommission_test.go
@@ -3,6 +3,7 @@ package cassandracluster
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"testing"
 
 	"github.com/jarcoal/httpmock"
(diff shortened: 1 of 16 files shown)


pkg/controller/cassandracluster/pod_operation.go
```diff
@@ -290,46 +299,43 @@ func (rcc *ReconcileCassandraCluster) ensureDecommission(cc *api.CassandraCluste
...
-			return breakResyncLoop, fmt.Errorf("For Status Ongoing we should have a PodLastOperation Pods item")
-		}
+	case api.StatusFinalizing:
 		lastPod, err := rcc.GetPod(cc.Namespace, podLastOperation.Pods[0])
 		if err != nil {
-			if !apierrors.IsNotFound(err) {
-				return breakResyncLoop, fmt.Errorf("failed to get last cassandra's pods '%s': %v",
-					podLastOperation.Pods[0], err)
+			if apierrors.IsNotFound(err) {
+				return rcc.deletePodPVC(cc, dcName, rackName, status, lastPod)
```
erdrix, line 316: is that what we want to do ? 
Was there a reason that we didn't remove it before ?

pkg/controller/cassandracluster/decommission_test.go
```diff
@@ -0,0 +1,216 @@
...
+	rcc.client.Update(context.TODO(), pdb)
+
+	return rcc, req
+}
+
+func registerJolokiaOperationModeResponder(host podName, op operationMode) {
+	fmt.Println(string(op))
+	httpmock.RegisterResponder("POST", JolokiaURL(host.FullName, jolokiaPort),
+		httpmock.NewStringResponder(200, fmt.Sprintf(`{"request":
+											{"mbean": "org.apache.cassandra.db:type=StorageService",
```
erdrix, line 48: indentation ?

pkg/controller/cassandracluster/pod.go
```diff
@@ -41,8 +41,8 @@ var reEndingNumber = regexp.MustCompile("[0-9]+$")
 // PodContainersReady returns true if all container in the Pod are ready
 func PodContainersReady(pod *v1.Pod) bool {
 	if pod.Status.ContainerStatuses != nil && len(pod.Status.ContainerStatuses) > 0 {
-		for _, c := range pod.Status.ContainerStatuses {
-			if c.Ready == false {
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if containerStatus.Ready == false {
```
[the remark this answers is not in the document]
cscetbon: right, I didn't look into that code but just renamed the variable. Doing the change

pkg/controller/cassandracluster/decommission_test.go
cscetbon: nope it's an inner call

pkg/controller/cassandracluster/pod_operation.go
cscetbon: I just changed when it's done to fix the problem

PR #238, opened 2020-07-13
Backup/Restore

cscetbon:
| Q               | A
| --------------- | ---
| New feature?    | no|yes

### What's in this PR?

Add Backup/Restore ability


Readme.md
```diff
@@ -35,7 +34,79 @@ For detailed installation instructions, see the [Casskop Documentation Page](htt
 
 The documentation of the Casskop operator project is available at the [Casskop Documentation Page](https://orange-opensource.github.io/casskop/docs/1_concepts/1_introduction).
 
-## Cassandra operator
+
+CassKop deals with Cassandra clusters on one datacenter. For multi-dacenters deployments, please use [MultiCassKop](multi-casskop/Readme.md) in addition to CassKop. This second operator is part of this same repository.
```
erdrix, line 38: This part wasn't removed ?

pkg/controller/cassandrarestore/reconcile.go
```diff
@@ -0,0 +1,308 @@
...
+	typeMeta := restore.TypeMeta
+	err := r.client.Update(ctx, restore)
+	if err != nil {
+		return nil, err
+	}
+	restore.TypeMeta = typeMeta
+	return restore, nil
+}
+
+func (r *ReconcileCassandraRestore) listPods(namespace string, selector map[string]string) (*v1.PodList, error) {
```
erdrix, line 295: this code is never re-used somewhere else ?

pkg/controller/common/common_controller.go
```diff
@@ -0,0 +1,40 @@
...
+	return reconcile.Result{}, err
+}
+
+// reconciled returns an empty result with nil error to signal a successful reconcile
+// to the controller manager
+func Reconciled() (reconcile.Result, error) {
+	return reconcile.Result{}, nil
+}
+
+// newSidecarsConnection is a convenience wrapper for creating a sidecars connection
```
erdrix, line 29: function renamed !

pkg/controller/cassandrarestore/reconcile.go
cscetbon: there are identical functions but on different reconcile object. Maybe we should refactor it under as they all use the client of the reconcile object ?

Readme.md
cscetbon: Correct, I don't know why it's removed. Adding it back

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
+| Q               | A
+| --------------- | ---
+| Bug fix?        | no|yes
+| New feature?    | no|yes
+| API breaks?     | no|yes
+| Deprecations?   | no|yes
```
cscetbon, line 6: I think we can replace them by checkboxes instead

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
+# How to contribute
+
+CassKop is Apache 2.0 licensed and accepts contributions via GitHub pull requests. This document outlines some of the
+conventions on commit message formatting, contact points for developers, and other resources to help get contributions
+into CassKop.
+
+## Email and Chat
+
```
cscetbon, line 8: There is no email. Should we keep only Slack ? I think it makes sense

Readme.md
```diff
@@ -1,14 +1,16 @@
+<img src="static/casskop.png" alt="Logo" width="150"/>
+
 # CassKop - Cassandra Kubernetes operator
 
-[![CircleCI](https://circleci.com/gh/Orange-OpenSource/cassandra-k8s-operator.svg?style=svg&circle-token=480ca5c31a9e9ef9b893151dd2d7c15eaf0e94d0)](https://circleci.com/gh/Orange-OpenSource/cassandra-k8s-operator)
+[![CircleCI](https://circleci.com/gh/Orange-OpenSource/casskop.svg?style=svg&circle-token=480ca5c31a9e9ef9b893151dd2d7c15eaf0e94d0)](https://circleci.com/gh/Orange-OpenSource/casskop) [![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=Orange-OpenSource_cassandra-k8s-operator&metric=alert_status)](https://sonarcloud.io/dashboard?id=Orange-OpenSource_cassandra-k8s-operator)
```
cscetbon, line 5: It's not a private repository. We shouldn't need a token

cassandra-stress/cassandra-stress-medium.yaml
```diff
@@ -14,18 +14,18 @@ spec:
...
   - name: cassie1-cassandra-stress
-    image: orangeopensource/cassandra-image
+    image: cassandra
     imagePullPolicy: "IfNotPresent"
     securityContext:
       capabilities:
         add: ["IPC_LOCK"]
     command: ["/bin/sh"]
-    args: ["-c", "cassandra-stress 'user profile=/opt/cassandra-stress/medium_stress.yaml ops(insert=1) n=5m cl=one -node cassandra-demo-dc1.cassandra-demo -mode native cql3 user=bench password=monbench -rate threads=50 -pop seq=0..5m -graph file=/tmp/stress-medium1.html' && echo END && while true ; do sleep 60; done"]
+    args: ["-c", "cassandra-stress 'user profile=/opt/cassandra-stress/medium_stress.yaml ops(insert=1,query_by_sub_id=1) n=5m cl=one -node cassandra-demo -mode native cql3 user=cassandra password=cassandra -rate threads=50 -pop seq=0..5m -graph file=/tmp/stress-medium1.html' && echo END && while true ; do sleep 60; done"]
```
cscetbon, line 28: why add `query_by_sub_id`? what about the other files ?

CONTRIBUTING.md
erdrix: You can add the e-mail : prj.casskop.support@list.orangeportails.net

pkg/controller/cassandrabackup/backup.go
```diff
@@ -0,0 +1,93 @@
...
+		corev1.EventTypeNormal,
+		"BackupInitiated",
+		fmt.Sprintf("Task initiated to backup datacenter %s of cluster %s to %s under snapshot %s",
+			backupClient.backup.Spec.Datacenter, backupClient.backup.Spec.CassandraCluster,
+			backupClient.backup.Spec.StorageLocation, backupClient.backup.Spec.SnapshotTag))
+
+	for range time.NewTicker(2 * time.Second).C {
+		if status, err := backrestClient.BackupStatusByID(operationID); err != nil {
+			logging.Error(err, fmt.Sprintf("Error while finding submitted backup operation %v", operationID))
+			break
```
erdrix, line 52: Do we really want to break the loop ? 
I mean if the error is due to the operation not found : ok,  but if it's just a timeout or something else, the next call might work, so why not try again ?

pkg/controller/cassandrabackup/backup.go
```diff
@@ -0,0 +1,93 @@
...
+		} else {
+			backupClient.updateStatus(status, logging)
```
erdrix, line 54: You might be in a case that the Backup update failed, and you only log the error, but the object will not be updated :/
Shouldn't we catch the error and try again on the next iteration ?

pkg/controller/cassandrabackup/reconcile.go
```diff
@@ -0,0 +1,249 @@
...
+
+		// Add Finalizer
+		r.setFinalizer(cassandraBackup, true)
+	}
+
+	if cassandraBackup.IsScheduled() {
+		return reconcile.Result{}, nil
+	}
+
+	return reconcile.Result{}, r.backupData(cassandraBackup, cc, reqLogger)
```
erdrix, line 228: What happens if we apply changes to a CassandraBackup and a backup is already running ? 
Is it starting another backup ? Does it kill the one who is running ? Does it do nothing ?

pkg/controller/cassandrabackup/backup.go
cscetbon: @erdrix I don't think it makes sense. Only 2 possible errors https://github.com/Orange-OpenSource/casskop/blob/backup-operator/pkg/cassandrabackup/operations.go#L55-L74 and both of them are not recoverable

pkg/controller/cassandrabackup/reconcile.go
cscetbon: @erdrix If the backup is scheduled, it's updated. If it's not, no new backup runs https://github.com/Orange-OpenSource/casskop/blob/backup-operator/pkg/controller/cassandrabackup/reconcile.go#L120-L121

erdrix (dismissed):

PR #252, opened 2020-09-01
[Fix/Plugin] Remove metadata.resourceVersion from the applied resource

### LINKED ISSUE #251: Kubectl apply fails after rolling restart triggered by plugin
## Bug Report

**What did you do?**
- We triggered a rolling restart with the casskop plugin:
  kubectl casskop restart --crd cassandra-bgl --full
- We applied the initial configuration of the cluster
  kubectl apply -f cassandra-bgl.yaml

Nothing is supposed to happen as we applied the initial yaml.
We are not supposed to have an error

We got the message:
The cassandraclusters "cassandra-bgl" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update

**Environment**
* casskop version:
0.5.4

1.16.8

* Kubernetes cluster kind: 
Orange internal cluster 

* Cassandra version:
3.11

**Possible Solution**
Similar problem on other operator:
https://github.com/argoproj/argo-cd/issues/3657#issuecomment-634933960

Seems not to come from the operator but likely the plugin

erdrix:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #251 

### What's in this PR?

This PR removes the `metadata.resourceVersion` field before applying the resource when the plugin is used.

### Why?

Without this, once the plugin has performed an operation that requires an `apply`, we aren't able to apply the changes without removing the `metadata.annotations.kubectl.kubernetes.io/last-applied-configuration` field.

### Checklist

- [X] Implementation tested
- [X] Logging code meets the guideline
- [X] User guide and development docs updated (if needed)
- [x] Append changelog

### DIFF
--- a/plugins/kubectl-casskop
+++ b/plugins/kubectl-casskop
@@ -30,6 +30,9 @@ RE_RACK_NAME = re.compile(r"(\w+)\W(\w+)")
 
 def k_apply_with_input(input, error, *args):
     try:
+        resource = json.loads(input)
+        del resource["metadata"]["resourceVersion"]
+        input = json.dumps(resource)
         check_output(["kubectl", "apply", "-f", "-"], stderr=STDOUT, input=str.encode(input))
     except CalledProcessError:
         die(error)
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,8 @@
 
 ### Bug Fixes
 
+- PR [#252](https://github.com/Orange-OpenSource/casskop/pull/252) - **[Plugin]** Remove metadata.resourceVersion from the applied resource
+
 
 ## 0.5.4
 


plugins/kubectl-casskop
```diff
@@ -30,6 +30,9 @@ RE_RACK_NAME = re.compile(r"(\w+)\W(\w+)")
 
 def k_apply_with_input(input, error, *args):
     try:
+        resource = json.loads(input)
+        del resource["metadata"]["resourceVersion"]
+        input = json.dumps(resource)
```
cscetbon, line 35: Can you add a unit test along with this change ?

PR #263, opened 2020-09-22
Unset RollingRestart flag before comparing CassandraCluster objects in MultiCasskop

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | []
| API breaks?     | []
| Deprecations?   | []

MultiCasskop prevents a rolling restart by rollbacking the change when it sees the flag RollingRestart being set

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,7 @@ orbs:
       vm:
         machine:
           image: ubuntu-1604:201903-01
+        resource_class: large
     jobs:
       e2e-orb:
         # PARAMETERS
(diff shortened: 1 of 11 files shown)


fdehay:
If I understand correctly, you are unsetting the flag RollingRestart in the spec of the CC.
Will this not prevent the Rolling Restart to occur at CassKop level?
May be you are working on a copy of the actual CC and then it would work. Just want to check :)

allamand:
There are also other fields that are removed for the comparison, I think you just need to add this one also

cscetbon:
> If I understand correctly, you are unsetting the flag RollingRestart in the spec of the CC.
> Will this not prevent the Rolling Restart to occur at CassKop level?
> May be you are working on a copy of the actual CC and then it would work. Just want to check :)

Yeah it's just an object used in the comparison to decide if there is a change to apply or not to underlying cassandracluster objects

cscetbon:
I still see some messages that shouldn't be there 
```
time="2020-09-27T18:18:56Z" level=debug msg="CassandraCluster does not exist, we create it " 
  cluster=cassandra-demo kubernetes=k8s-cluster2 namespace=cluster2
time="2020-09-27T18:18:56Z" level=debug msg="Create CassandraCluster" cluster=cassandra-demo 
  kubernetes=k8s-cluster2 namespace=cluster2
time="2020-09-27T18:18:56Z" level=info msg="Just Update CassandraCluster, returning for now.." 
  cluster=cassandra-demo kubernetes=k8s-cluster2 namespace=cluster2
```

It's the last thing I have to figure out before this PR is ready for review

fdehay (approved):
Ok for me

PR #273, opened 2020-10-20
Fix joloika auth

### LINKED ISSUE #270: Setting imageJolokiaSecret does not enable auth
**What did you do?**
I configured `imageJolokiaSecret`, but Jolokia started up without auth.

This issue was previously reported in #211, but it was closed before confirming whether or not this was really a problem. I can confirm that this is a bug in need of fixing.

Secret
```
apiVersion: v1
kind: Secret
metadata:
  name: jolokia-auth
type: Opaque
data:
  username: YWRtaW4=
  password: cGFzc3dvcmQK
```

CassandraCluster
```
apiVersion: "db.orange.com/v1alpha1"
kind: "CassandraCluster"
metadata:
  name: cassandra-demo
  labels:
    cluster: k8s.kaas
spec:
  cassandraImage: cassandra:3.11
  bootstrapImage: orangeopensource/cassandra-bootstrap:0.1.5
  configMapName: cassandra-configmap-v1
  dataCapacity: "200Mi"
  dataStorageClass: "standard"
  imagepullpolicy: IfNotPresent
  hardAntiAffinity: false
  deletePVC: true
  autoPilot: false
  gcStdout: false
  autoUpdateSeedList: false
  maxPodUnavailable: 1
  imageJolokiaSecret:
    name: jolokia-auth
  runAsUser: 999
  resources:
    requests:
      cpu: '1'
      memory: 200Mi
    limits:
      cpu: '1'
      memory: 200Mi
  topology:
    dc:
      - name: dc1
        nodesPerRacks: 1
        rack:
          - name: rack1
```

I expected to see Cassandra start up with this jvm option:
```
-javaagent:/extra-lib/jolokia-agent.jar=host=0.0.0.0,executor=fixed,authMode=basic,user=admin,password=password
```

Cassandra started up with this jvm option:
```
-javaagent:/extra-lib/jolokia-agent.jar=host=0.0.0.0,executor=fixed
```

You can see that the Jolokia env vars are set in the cassandra container:
```
$ kubectl -n cassandra exec -it cassandra-demo-dc1-rack1-0 -c cassandra -- /bin/bash
cassandra@cassandra-demo-dc1-rack1-0:/$ env | sort | grep JOLOKIA
CASSANDRA_AUTH_JOLOKIA=true
JOLOKIA_PASSWORD=password
JOLOKIA_USER=admin
```

However, these env vars are not set in the bootstrap container, which is where `run.sh` looks for them in order to update the jolokia jvm option. If you add a `sleep 3600` to `post_run.sh` and exec into the bootstrap pod, you can see that the JOLOKIA env vars are missing:
```
$ kubectl -n cassandra exec -it cassandra-demo-dc1-rack1-0 -c bootstrap -- /bin/bash
cassandra@cassandra-demo-dc1-rack1-0:/$ env | sort | grep JOLOKIA
JOLOKIA_VERSION=1.6.1
```

Bootstrap env vars are set in `cassandracluster.bootstrapContainerEnvVar()`, but the Jolokia env vars are not included in the list of env vars set.
https://github.com/Orange-OpenSource/casskop/blob/9253cd6115accbd715007817cc4b78b0c2865866/pkg/controller/cassandracluster/generator.go#L577

The Jolokia env vars are added to the cassandra container (and only the cassandra container) here:
https://github.com/Orange-OpenSource/casskop/blob/9253cd6115accbd715007817cc4b78b0c2865866/pkg/controller/cassandracluster/generator.go#L381-L408

**Environment**
* casskop version:

```
v0.5.6-release
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T21:51:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1", GitCommit:"206bcadf021e76c27513500ca24182692aabd17e", GitTreeState:"clean", BuildDate:"2020-09-14T07:30:52Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
```
* Kubernetes cluster kind:
```
$ kind version
kind v0.9.0 go1.15.2 darwin/amd64
```
* Cassandra version:

```
3.11.8
```

**Possible Solution**

For the Jolokia credentials in particular, I'd like the JOLOKIA_USER and JOLOKIA_PASSWORD env vars to be set in both the bootstrap and cassandra containers.

In general, I don't see any reason why the env vars in the bootstrap and cassandra containers should differ. Is there a reason why they shouldn't be the same?

toffer:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]
| New feature?    | []
| API breaks?     | []
| Deprecations?   | []
| Related tickets | fixes #270

### What's in this PR?
This PR adds JOLOKIA env vars to the bootstrap container.

The JOLOKIA env vars must be set in the bootstrap container in order for run.sh to enable auth for Jolokia. And, they also need to be set in the cassandra container, so that readiness-probe.sh can use the credentials when required.

### Additional context
I'd like to add a unit test to check for JOLOKIA env vars, but I'm not sure how to set a `jolokia-auth` secret in the unit tests. Happy to add that if you can give me guidance on how I could do that.

I did confirm that unit tests pass, and the env vars are set as expected in both the bootstrap and cassandra containers when running the controller.

### Checklist

- [x] Implementation tested
- [x] Append changelog

### DIFF
--- a/docker/bootstrap/files/readiness-probe.sh
+++ b/docker/bootstrap/files/readiness-probe.sh
@@ -14,8 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Add Jolokia credentials, if defined in environment.
+if [[ -z $JOLOKIA_USER ]] || [[ -z $JOLOKIA_PASSWORD ]]; then
+    USER_OPT=""
+else
+    USER_OPT="--user $JOLOKIA_USER:$JOLOKIA_PASSWORD"
+fi
+
 # We check when the node is up and in normal state
-CURL="/opt/bin/curl -s --connect-timeout 0.5"
+CURL="/opt/bin/curl $USER_OPT -s --connect-timeout 0.5"
 BASE_CMD="http://$POD_IP:8778/jolokia/read/org.apache.cassandra.db:type=StorageService"
 
 if $CURL ${BASE_CMD}/LiveNodes | grep -q $POD_IP; then
--- a/docker/bootstrap/files/readiness-probe.sh
+++ b/docker/bootstrap/files/readiness-probe.sh
@@ -15,9 +15,7 @@
 # limitations under the License.
 
 # Add Jolokia credentials, if defined in environment.
-if [[ -z $JOLOKIA_USER ]] || [[ -z $JOLOKIA_PASSWORD ]]; then
-    USER_OPT=""
-else
+if [[ -n $JOLOKIA_USER ]] && [[ -n $JOLOKIA_PASSWORD ]]; then
     USER_OPT="--user $JOLOKIA_USER:$JOLOKIA_PASSWORD"
 fi
 
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -596,10 +596,6 @@ func bootstrapContainerEnvVar(cc *api.CassandraCluster, status *api.CassandraClu
 	return bootstrapEnvVars
 }
 
-func cassandraContainerEnvVar(cc *api.CassandraCluster) []v1.EnvVar {
-	return commonBootstrapCassandraEnvVar(cc)
-}
-
 func commonBootstrapCassandraEnvVar(cc *api.CassandraCluster) []v1.EnvVar {
 	commonEnvVars := []v1.EnvVar{
 		{
@@ -768,7 +764,7 @@ func createCassandraContainer(cc *api.CassandraCluster, status *api.CassandraClu
 				},
 			},
 		},
-		Env: cassandraContainerEnvVar(cc),
+		Env: commonBootstrapCassandraEnvVar(cc),
 		ReadinessProbe: &v1.Probe{
 			InitialDelaySeconds: *cc.Spec.ReadinessInitialDelaySeconds,
 			TimeoutSeconds:      *cc.Spec.ReadinessHealthCheckTimeout,
--- a/pkg/controller/cassandracluster/generator.go
+++ b/pkg/controller/cassandracluster/generator.go
@@ -378,35 +378,6 @@ func generateCassandraStatefulSet(cc *api.CassandraCluster, status *api.Cassandr
 		}
 	}
 
-	for idx, container := range ss.Spec.Template.Spec.Containers {
-		if container.Name == cassandraContainerName {
-			if (cc.Spec.ImageJolokiaSecret != v1.LocalObjectReference{}) {
-				ss.Spec.Template.Spec.Containers[idx].Env = append(container.Env,
-					v1.EnvVar{
-						Name: "JOLOKIA_USER",
-						ValueFrom: &v1.EnvVarSource{
-							SecretKeyRef: &v1.SecretKeySelector{
-								LocalObjectReference: cc.Spec.ImageJolokiaSecret,
-								Key:                  "username",
-							},
-						},
-					},
-					v1.EnvVar{
-						Name: "JOLOKIA_PASSWORD",
-						ValueFrom: &v1.EnvVarSource{
-							SecretKeyRef: &v1.SecretKeySelector{
-								LocalObjectReference: cc.Spec.ImageJolokiaSecret,
-								Key:                  "password",
-							},
-						},
-					},
-					v1.EnvVar{
-						Name:  "CASSANDRA_AUTH_JOLOKIA",
-						Value: "true"})
-			}
-		}
-	}
-
 	// Merge cassandra main container environment variables into sidecars.
 
 	var sidecarEnv []v1.EnvVar
@@ -580,8 +551,7 @@ func bootstrapContainerEnvVar(cc *api.CassandraCluster, status *api.CassandraClu
 	//in statefulset.go we surcharge this value with conditions
 	seedList := cc.SeedList(&status.SeedList)
 	numTokensPerRacks := cc.NumTokensPerRacks(dcRackName)
-
-	return []v1.EnvVar{
+	bootstrapEnvVars := []v1.EnvVar{
 		{
 			Name:  "CASSANDRA_MAX_HEAP",
 			Value: defineJvmMemory(resources).maxHeapSize,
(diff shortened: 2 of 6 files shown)


toffer:
Should I be concerned about the 9 checks that seem stuck in "Waiting for status"?

allamand:
Maybe you. Can see tests for bootstrap in docker/bootstrap/dgoss test suit also ?

toffer:
Looks like the e2e tests are being skipped by design. Just noticed these filters in the circleci config: `canonical_repo_filter` and `ignore_fork_pr_filter`.

https://github.com/Orange-OpenSource/casskop/blob/master/.circleci/config.yml#L348-L362

docker/bootstrap/files/readiness-probe.sh
```diff
@@ -14,8 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Add Jolokia credentials, if defined in environment.
+if [[ -z $JOLOKIA_USER ]] || [[ -z $JOLOKIA_PASSWORD ]]; then
+    USER_OPT=""
+else
+    USER_OPT="--user $JOLOKIA_USER:$JOLOKIA_PASSWORD"
+fi
```
cscetbon, line 22:
```suggestion
if [[ -n $JOLOKIA_USER ]] && [[ -n $JOLOKIA_PASSWORD ]]; then
    USER_OPT="--user $JOLOKIA_USER:$JOLOKIA_PASSWORD"
fi
```

cscetbon: You don't really need an if/else cause it's shell scripting, if the variable doesn't exist it will be empty

pkg/controller/cassandracluster/generator.go
```diff
@@ -630,6 +591,55 @@ func bootstrapContainerEnvVar(cc *api.CassandraCluster, status *api.CassandraClu
 			},
 		},
 	}
+	commonEnvVars := commonBootstrapCassandraEnvVar(cc)
+	bootstrapEnvVars = append(bootstrapEnvVars, commonEnvVars...)
+	return bootstrapEnvVars
+}
+
+func cassandraContainerEnvVar(cc *api.CassandraCluster) []v1.EnvVar {
```
cscetbon, line 599: You can get rid of that function and just call `commonBootstrapCassandraEnvVar(cc)` at https://github.com/Orange-OpenSource/casskop/pull/273/files#diff-8f39a9e34ee5c81ededc0c74d2d2e8ffc5dece79eab1cd4223461aa073b3a9c4R771 like you did above

pkg/controller/cassandracluster/generator_test.go
```diff
@@ -499,6 +499,18 @@ func checkVarEnv(t *testing.T, containers []v1.Container, cc *api.CassandraClust
...
+			podIP := v1.EnvVar{
+				Name: "POD_IP",
+				ValueFrom: &v1.EnvVarSource{
+					FieldRef: &v1.ObjectFieldSelector{
+						APIVersion: "v1",
+						FieldPath:  "status.podIP",
+					},
+				},
+			}
+			assert.Contains(container.Env, podIP)
```
cscetbon, line 513: You're checking that cassandra container has the POD_IP env, but what about the common env vars between cassandra and bootstrap ?

cscetbon:
@allamand you remember why we don't run e2e tests on PRs ?
@toffer I left a few comments/suggestion changes.

docker/bootstrap/files/readiness-probe.sh
toffer: 👍 I'll update.

pkg/controller/cassandracluster/generator.go
toffer: It's probably premature, but I was trying to make it clear which variables get defined where: cassandra-only env vars get defined in `cassandraContainerEnvVar()`, bootstrap-only env vars get defined `bootstrapContainerEnvVar()`, and the common ones go in a common function.

Since the cassandra container ONLY contains common variables right now, it's not really necessary, but I was trying to make it obvious where to define cassandra-only env vars in the future.

I'll scratch that idea and just call `commonBootstrapCassandraEnvVar()` directly.

pkg/controller/cassandracluster/generator_test.go
toffer: The only other env vars in the cassandra container are the jolokia-related ones: JOLOKIA_USER, JOLOKIA_PASSWORD, and CASSANDRA_AUTH_JOLOKIA.

As I mentioned above, I'd love to test for the presence for these, but that requires setting a `jolokia-auth` secret so this `cassandracluster` configuration works:
```
  imageJolokiaSecret:
    name: jolokia-auth
```
I'm not sure how to set a secret in the unit tests. Help?

toffer:
@cscetbon Thanks! I incorporated your feedback, so all that's left is the testing question. I'll need your guidance there.

allamand:
We don't run e2e in pr to prevent leak of secrets used in the pipeline.

We have 2 kind of tests: the one which test the docket image and their entry-points using dgoss

Oran's the other ones which really deploy cassandra in a test environment and check that scenarios ara able to work as attended.

The problem with e2e, is that it's very long tests that spin up k8s and cassandra so it take more than 30mn actually.. so we need to add more tests with care

cscetbon:
Agreed, and I found the patten we followed https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks. I triggered the tests

cscetbon:
@toffer e2e tests are failing because you changed the bootstrap image and we need to publish a new one. So please increment the version at https://github.com/Orange-OpenSource/casskop/blob/master/docker/bootstrap/Makefile#L22 and I'll then publish a new version for the test to use it and pass with the new version

pkg/controller/cassandracluster/generator_test.go
cscetbon: @toffer no need it's already used in e2e tests see https://github.com/Orange-OpenSource/casskop/blob/master/test/e2e/testdata/cassandracluster-1DC.yaml#L19-L20

toffer:
@cscetbon I bumped the bootstrap version to `0.1.6`.

pkg/controller/cassandracluster/generator_test.go
toffer: @cscetbon I see. These e2e tests can't pass if the readiness-probe.sh scripts fail, and the readiness-probe.sh script can't pass if the JOLOKIA env vars aren't set in the cassandra container.

Sounds good to me!

cscetbon:
@toffer yeah please also change it at https://github.com/Orange-OpenSource/casskop/blob/master/pkg/apis/db/v1alpha1/cassandracluster_types.go#L38. I forgot that one too

cscetbon (approved):

PR #276, opened 2020-11-04
Upgrade Icarus and add errors to status

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]

### What's in this PR?

Some changes were made on Icarus to send the bubble up errors that could happen on any sidecar in a global request.

### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here


pkg/apis/db/v1alpha1/cassandrabackup_types.go
```diff
@@ -47,21 +48,24 @@ type CassandraBackupSpec struct {
...
 }
 
-type BackupState string
+type BackupConditionType string
 
 const (
-	BackupRunning   BackupState = "RUNNING"
-	BackupCompleted BackupState = "COMPLETED"
-	BackupFailed    BackupState = "FAILED"
+	BackupRunning   BackupConditionType = "RUNNING"
```
smiklosovic, line 54: @cscetbon all possible states an operation can be in: https://github.com/instaclustr/instaclustr-commons/blob/master/src/main/java/com/instaclustr/operations/Operation.java#L175

cscetbon: Yes @smiklosovic  but we're not using the other types in our tests. In our use, `CANCELLED` should never happen, you confirm ?

smiklosovic: yeah it shouldnt occur in your case.

erdrix (approved):

PR #278, opened 2020-11-12
liveness - Make separate liveness/readiness probes possible.

### LINKED ISSUE #277: Make separate liveness/readiness probes possible
## Feature Request

**Is your feature request related to a problem? Please describe.**

I'd like to make it possible to have different scripts for the liveness and readiness probes. As of #234, the `LivenessProbe` command in the cassandra container `PodTemplateSpec` is hard-coded to point to `/etc/readiness-probe.sh`, so there's no way to have separate scripts.

**Describe the solution you'd like to see**

If we add `liveness-probe.sh` to the bootstrap Docker image, but make it a symlink to `readiness-probe.sh`, then we can keep the current behavior, where we use the same script for both probes. And, this also makes it possible to override this behavior and define a separate script for each probe.

**Describe alternatives you've considered**

Rather than use a symlink, we could cut/paste the contents of `readiness-probe.sh` into `liveness-probe.sh`. But, using the symlink makes it more obvious that we are using the same script for both probes (in the default case).

toffer:
| Q               | A
| --------------- | ---
| Bug fix?        | []
| New feature?    | [x]
| API breaks?     | []
| Deprecations?   | []
| Related tickets | fixes #277

### What's in this PR?
This PR adds back the ability to use different scripts for the liveness and readiness probes, but it preserves the default behavior (introduced in #234) where the same script is used for both.

The `readiness-probe.sh` script will continue to be used for both the liveness and readiness probes, since `liveness-probe.sh` is a symlink to `readiness-probe.sh`.

However, by using separate filenames for each of the probes, it is now possible to override this behavior, and define separate scripts for each type of probe.

### Checklist

- [x] Implementation tested
- [ ] Append changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ DOCKER_REPO_BASE_TEST ?= orangeopensource
 IMAGE_NAME := $(SERVICE_NAME)
 BUILD_IMAGE ?= orangeopensource/casskop-build
 
-BOOTSTRAP_IMAGE ?= orangeopensource/cassandra-bootstrap:0.1.4
+BOOTSTRAP_IMAGE ?= orangeopensource/cassandra-bootstrap:0.1.7
 TELEPRESENCE_REGISTRY ?= datawire
 KUBESQUASH_REGISTRY:=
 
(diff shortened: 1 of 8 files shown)


toffer:
For the dgoss test to pass, this will require building a new bootstrap docker image (version 0.1.7).

cscetbon:
I wanted to simplify it but I think it makes sense if you want to overwrite it. I’ll generate the bootstrap image and kick off he tests again

cscetbon (approved):

PR #287, opened 2020-12-17
Restrict psp role in helm chart to get and list

fdehay:
| Q               | A
| --------------- | ---
| Bug fix?        | []
| New feature?    | []
| API breaks?     | []
| Deprecations?   | []

### What's in this PR?
This Pr restricts the pod security policies that need to be applied from "use" to "list" and "watch"

### Why?
In our internal K8S cluster the "use" policy is not allowed at it enables to use the "priviledged" policy

### Checklist
- [X] Implementation tested

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,10 +106,10 @@ orbs:
                     false;
                   }
                 fi
-#                if [ << parameters.test_type >> = "kuttl" ]
-#                then
-#                  make kuttl-test-fix-arg << parameters.test_name >>
-#                fi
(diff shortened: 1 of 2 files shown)


PERES-Richard:
I think we should test the local (modified) chart with Kuttl, like we did on the previous PRs before the 1.0.2. Just uncomment this section to try out :

https://github.com/Orange-OpenSource/casskop/blob/master/.circleci/config.yml#L109

PERES-Richard (approved):

cscetbon (approved):

PR #289, opened 2021-01-15
Do not use data folder to store downloaded sstables

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]

There are some issues with the backup when downloadedsstables folders is put in data

### DIFF
--- a/pkg/backrest/backrest.go
+++ b/pkg/backrest/backrest.go
@@ -41,7 +41,7 @@ func (c *Client) PerformRestore(restore *api.CassandraRestore,
 		GlobalRequest: true,
 		Import_: &icarus.AllOfRestoreOperationRequestImport_{
 			Type_: "import",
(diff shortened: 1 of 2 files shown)


PERES-Richard:
Found this, don't know if it's need a change (maybe also for PR #290 since there is "HARDLINKS3 ?)

 https://github.com/Orange-OpenSource/casskop/blob/86e226ac592264df014ec411fbc60ab8260d904a/pkg/cassandrabackup/operations_test.go#L46

cscetbon:
> Found this, don't know if it's need a change (maybe also for PR #290 since there is "HARDLINKS3 ?)
> 
> https://github.com/Orange-OpenSource/casskop/blob/86e226ac592264df014ec411fbc60ab8260d904a/pkg/cassandrabackup/operations_test.go#L46

done @PERES-Richard

PERES-Richard (approved):

PR #290, opened 2021-01-15
Enforce HARDLINKS restore strategy as it is faster

cscetbon:
IMPORT restore strategy implies reading sstables and rewriting them which is longer that using hardlinks. That PR enforces that choice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,7 @@ docker-generate-k8s:
 		-v $(shell go env GOCACHE):/root/.cache/go-build:delegated --env GO111MODULE=on \
 		--env https_proxy=$(https_proxy) --env http_proxy=$(http_proxy) \
 		$(BUILD_IMAGE):$(OPERATOR_SDK_VERSION) /bin/bash -c 'operator-sdk generate k8s'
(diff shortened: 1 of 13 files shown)


PERES-Richard (changes requested):
I think this should be also removed : 
https://github.com/Orange-OpenSource/casskop/blob/7fc9abf7ea8fc0c950b773901e6a6e972c8b4884/pkg/apis/db/v1alpha1/common/test.go#L28

And also the doc should be updated as well
https://github.com/Orange-OpenSource/casskop/blob/restore-strategy/website/docs/6_references/6_cassandra_restore.md

(in YAML example & table row)

cscetbon:
@PERES-Richard I agree on the doc but not on the test. It's the response returned by the client and it always include the restore strategy

PERES-Richard:
> @PERES-Richard I agree on the doc but not on the test. It's the response returned by the client and it always includes the restore strategy

Okey. I will test this new implem on our cluster to strongly assert that this will not break backup & restore flows :)

PERES-Richard (approved):

PR #291, opened 2021-01-19
Add check if cassandraBackup.Annotations map is nil before assignement

PERES-Richard:
| Q               | A
| --------------- | ---
| Bug fix?        | [x]

### What's in this PR?
Add a map is nil check before setting last applied action on CassandraBackup.

### Why?
To avoid error when there's no annotation(s) (and so the map is nil).

### Checklist

- [ ] Append changelog

### DIFF
--- a/pkg/controller/cassandrabackup/reconcile.go
+++ b/pkg/controller/cassandrabackup/reconcile.go
@@ -106,7 +106,7 @@ func (r *ReconcileCassandraBackup) Reconcile(request reconcile.Request) (reconci
 		instanceChanged = false
 	} else {
 		if cassandraBackup.Annotations == nil {
-			cassandraBackup.Annotations = make(map[string] string )
+			cassandraBackup.Annotations = make(map[string] string)
 		}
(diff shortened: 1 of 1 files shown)


pkg/controller/cassandrabackup/reconcile.go
```diff
@@ -105,6 +105,9 @@ func (r *ReconcileCassandraBackup) Reconcile(request reconcile.Request) (reconci
 	if lac, _ := cassandraBackup.ComputeLastAppliedAnnotation(); lac == cassandraBackup.Annotations[annotationLastApplied] {
 		instanceChanged = false
 	} else {
+		if cassandraBackup.Annotations == nil {
+			cassandraBackup.Annotations = make(map[string] string )
```
cscetbon, line 109: can you remove the last extra space ?

PERES-Richard: Ofc ! Done

cscetbon (approved):

PR #294, opened 2021-02-02
Fix makefile crds generation

PERES-Richard:
| Q               | A
| --------------- | ---
| Bug fix        | [x]

### What's in this PR?
Correct CRDs generation rule

### Why?
To deploy CRDs in both deploy directory and in Casskop and Multi Casskop Helm Crds directories.

### Checklist
- [x] Implementation tested
- [ ] Append changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ docker-generate-crds:
 		--env https_proxy=$(https_proxy) --env http_proxy=$(http_proxy) \
 		$(BUILD_IMAGE):$(OPERATOR_SDK_VERSION) /bin/bash -c 'operator-sdk generate crds'
 	@sed -i '/\- protocol/d' deploy/crds/db.orange.com_cassandraclusters_crd.yaml
-	cp deploy/crds/* */helm/*/crds/
+	cp -v deploy/crds/* helm/*/crds/ && cp -v deploy/crds/* */helm/*/crds/
 
 docker-build-operator:
 	echo "Build Cassandra Operator. Using cache from "$(shell go env GOCACHE)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -185,7 +185,7 @@ docker-generate-k8s:
 		-v $(shell go env GOCACHE):/root/.cache/go-build:delegated --env GO111MODULE=on \
 		--env https_proxy=$(https_proxy) --env http_proxy=$(http_proxy) \
 		$(BUILD_IMAGE):$(OPERATOR_SDK_VERSION) /bin/bash -c 'operator-sdk generate k8s'
-	cp deploy/crds/* helm/*/crds/
+	cp deploy/crds/* */helm/*/crds/
```
cscetbon, line 188: generate-k8s generates only zz files, so I think we can get rid of it here

PERES-Richard: yeah I agree, done :)

Makefile
```diff
@@ -194,6 +194,7 @@ docker-generate-crds:
 		--env https_proxy=$(https_proxy) --env http_proxy=$(http_proxy) \
 		$(BUILD_IMAGE):$(OPERATOR_SDK_VERSION) /bin/bash -c 'operator-sdk generate crds'
 	@sed -i '/\- protocol/d' deploy/crds/db.orange.com_cassandraclusters_crd.yaml
+	cp deploy/crds/* */helm/*/crds/
```
cscetbon, line 196: here they are copied to multi-casskop but no longer to helm/ (at the root level) 👎

cscetbon (changes requested):
see my comments

Makefile
PERES-Richard: Fixed by duplicating the `cp ` to hit both `*/hem` et `./helm/` dir

cscetbon (approved):

cscetbon (dismissed):

cscetbon (approved):

PR #302, opened 2021-02-21
Bootstrap issue

### LINKED ISSUE #301: Time between native protocol is enabled and node is started matters
## Bug Report

When a node is live it's considered as ready and the next node is started, however the bootstrap of the next node will fail cause native protocol is not enable yet. 

**What did you do?**
Create a new cluster with 2 nodes and check if nodes see each other or not. It happens that they belong to different rings.

the bootstrap image tries to access the native port of seed nodes to know if they can join an existing ring. But as a node is ready before it can happen that they don't see each other.

nodes do not see each other

**Environment**
* casskop version: 1.1.2

**Possible Solution**
Test the JMX port instead of the native port.

cscetbon:
| Q               | A
| --------------- | ---
| Bug fix?        | [X]
| Related tickets | fixes #301 

### What's in this PR?

Test Jolokia port instead of native port when attempting to bootstrap nodes

### Why?

 It prevents having nodes being created in separated clusters


samples/cassandracluster.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   bootstrapImage: orangeopensource/cassandra-bootstrap:0.1.4
```
fdehay, line 9: why not 0.1.8?

samples/cassandracluster.yaml
```diff
@@ -9,7 +9,7 @@ spec:
...
   configMapName: cassandra-configmap-v1
   dataCapacity: "200Mi"
-  dataStorageClass: "local-storage"
+  dataStorageClass: local-path
```
fdehay, line 12: WHere does this local-path come from? Is it an old update?

fdehay:
Will CCL need to change their network policies for this to work on Rick?
Or is port 9042 open as well already?

samples/cassandracluster.yaml
cscetbon: K3d

cscetbon:
> Will CCL need to change their network policies for this to work on Rick?
> Or is port 9042 open as well already?
They’ll have to update their policies

fdehay (dismissed):

PR #306, opened 2021-03-02
Custom script when Cassandra starts up

SuhailRahman:
| Q               | A
| --------------- | ---
| New feature?    | [X]

### What's in this PR?
 This PR has the extended functionality of running a custom script when the Cassandra pod is started up, using the postStart lifecycle hook. One use case is when the user wishes to create custom roles on the startup of Cassandra
### Why?
In case a user wishes to perform an operation immediately after Cassandra is started up, such as create a role with cqlsh, the script specified by /etc/cassandra_post_start.sh can be injected via configmap to run on post-start.

### DIFF
--- a/tools/publish_helm_gcs.sh
+++ b/tools/publish_helm_gcs.sh
@@ -50,7 +50,6 @@ setup_helm_client() {
 
     PATH="$(pwd)/tmp/linux-amd64/:$PATH"
 
-    helm init --client-only
     helm repo add incubator-orange "$INCUBATOR_REPO_URL"
 }
 


ajaybhat:
@cscetbon can you please review this

pkg/controller/cassandracluster/generator.go
```diff
@@ -752,6 +752,15 @@ func createCassandraContainer(cc *api.CassandraCluster, status *api.CassandraClu
 		},
 
 		Lifecycle: &v1.Lifecycle{
+		    PostStart: &v1.Handler{
+				Exec: &v1.ExecAction{
+					Command: []string{
+						"/bin/bash",
+						"-c",
+						"/etc/cassandra/cassandra_post_start.sh",
```
cscetbon, line 760: Your changes would make any deployment that doesn't provide that script fail. Your PostStart action should check if the file exists and if it does should run.

Also it doesn't mean that cassandra was started but just that the container was started as it's mentioned in the documentation _"There is no guarantee, however, that the postStart handler is called before the Container's entrypoint is called"_
So I think there is still some work missing here

cscetbon (changes requested):

pkg/controller/cassandracluster/generator.go
cscetbon: @ajaybhat @SuhailRahman 👆

cscetbon:
Closing the PR for now as it's been opened for more than a month. Reopen it when you have more time to work on it. Thanks

PR #313, opened 2021-04-08
Multicasskop annotations

cscetbon:
| Q               | A
| --------------- | ---
| New feature?    | [X]

### What's in this PR?
Allow annotations to be passed when installing multicasskop chart


fdehay:
helm seems ok to me.
But why removing the 2 crds? Some cleanup?

cscetbon:
> helm seems ok to me.
> But why removing the 2 crds? Some cleanup?

Those are not CRDs but just objects. They shouldn’t be there

PR #340, opened 2021-07-07
Fix operator & pods restart when scaling up DC with autoUpdateSeedList set to true

PERES-Richard:
Signed-off-by: Richard Peres <richard.peres@orange.com>

| Q               | A
| --------------- | ---
| Bug fix        | [x]

### What's in this PR?
A small fix to a bug that appears when scaling up a CassandraCluster by adding a new DC. The bug causes other rack's pods to restart as well as operator's pod only when `autoUpdateSeedList` is set to true in cluster config.

### Checklist

- [x] Implementation tested
- [ ] Append changelog


PERES-Richard:
Maybe we will have to add a AssertCommand test step to ensure there were 0 restart from operator and others pods with something like `[[kgp -o json -o json | jq '.items[].status.containerStatuses[].restartCount] -eq 0]'`

PERES-Richard:
`k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
casskop/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x15d9800, 0x25a2050)
/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/Orange-OpenSource/casskop/pkg/controller/cassandracluster.FlipCassandraClusterUpdateSeedListStatus(0xc000100700, 0xc0002b8960)
casskop/pkg/controller/cassandracluster/reconcile.go:642 +0x192
github.com/Orange-OpenSource/casskop/pkg/controller/cassandracluster.(*ReconcileCassandraCluster).Reconcile(0xc000328c90, 0xc00041e080, 0x7, 0xc00041e070, 0xd, 0x0, 0x0, 0x0, 0x0)
casskop/pkg/controller/cassandracluster/cassandracluster_controller.go:150 +0x46c
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0001b29c0, 0x1637880, 0xc0006ffda0, 0x1733c00)
casskop/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:256 +0x161
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0001b29c0, 0x203000)
casskop/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232 +0xae
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker(0xc0001b29c0)
casskop/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211 +0x2b
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc000974140)
casskop/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155 +0x5f
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc000974140, 0x19df000, 0xc00096a240, 0xc00011a001, 0xc000934180)
casskop/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156 +0xa3
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000974140, 0x3b9aca00, 0x0, 0x185a101, 0xc000934180)
casskop/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0xe2
k8s.io/apimachinery/pkg/util/wait.Until(0xc000974140, 0x3b9aca00, 0xc000934180)
casskop/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:90 +0x4d
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
casskop/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:193 +0x305
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x14123d2]`

pkg/controller/cassandracluster/reconcile.go
```diff
@@ -640,6 +641,10 @@ func FlipCassandraClusterUpdateSeedListStatus(cc *api.CassandraCluster, status *
 				dcRackName := cc.GetDCRackName(dcName, rackName)
 				dcRackStatus := status.CassandraRackStatus[dcRackName]
 
+				if dcRackStatus == nil {
```
cscetbon, line 644: which one is actually failing ? this one or the latter ? Do we know why it happens ? It seems we are solving it down front but something happens up front ...

pkg/controller/cassandracluster/reconcile.go
```diff
@@ -659,8 +664,13 @@ func FlipCassandraClusterUpdateSeedListStatus(cc *api.CassandraCluster, status *
 
 					rackName := cc.GetRackName(dc, rack)
 					dcRackName := cc.GetDCRackName(dcName, rackName)
+
 					dcRackStatus := status.CassandraRackStatus[dcRackName]
 
+					if dcRackStatus == nil {
+						return errors.New("DC Rack Status empty")
+					}
```
cscetbon, line 672: This one is not needed. If the first test errors out that code won't be executed.

pkg/controller/cassandracluster/reconcile_test.go
```diff
@@ -85,7 +85,9 @@ func TestFlipCassandraClusterUpdateSeedListStatusScaleDC2(t *testing.T) {
 	UpdateCassandraClusterStatusPhase(cc, status)
 
 	//Flip with AutoUpdateSeedList= true -> update status
-	FlipCassandraClusterUpdateSeedListStatus(cc, status)
+	if err := FlipCassandraClusterUpdateSeedListStatus(cc, status); err != nil {
+		assert.Fail(err.Error())
+	}
```
cscetbon, line 90: All those checks are wrong IMO. Here it seems we don't know if it will happen or not, meaning the test is not deterministic which is not good. It's a unit test, it must be called or not but we need to know it.

test/kuttl/multi-dcs/00-createCluster.yaml
```diff
@@ -9,7 +18,7 @@ spec:
   hardAntiAffinity: false
   deletePVC: true
   autoPilot: true
-  autoUpdateSeedList: false
+  autoUpdateSeedList: true
   dataStorageClass: local-path
   resources:
     requests: &requests
```
cscetbon, line 24: whenever you see 
```
  resources:
    requests: &requests
      memory: ...
      cpu: ...
    limits: *requests
```
please replace it by
```
  resources:
    limits:
      memory: ...
      cpu: ...
```

test/kuttl/multi-dcs/01-assert.yaml
```diff
@@ -1,3 +1,10 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 500
+commands:
+  - script: "kubectl get pods -n $NAMESPACE -o json | jq '.items | length' | jq 'if . != 5 then halt_error(-2) else \"OK\" end' 2> /dev/null"
```
cscetbon, line 5: You can chain calls in jq:
```
$ pbpaste|jq 'if(.topping|length) > 5 then "ok" else "nok" end' 
"ok"
```

test/kuttl/multi-dcs/02-unallowedScaleDown.yaml
```diff
@@ -8,7 +8,9 @@ spec:
       - name: dc1
         rack:
           - name: rack1
+          - name: rack2
```
cscetbon, line 11: I thought you could reproduce it with 2 DCs of 1 rack ?

cscetbon (changes requested):

pkg/controller/cassandracluster/reconcile.go
PERES-Richard: Both actually.. When I have fixed the 1st dcRackStatus assignment, which cause the operator to crash as logs said, the operator still crashed because of the second. It's like the value of cc & status in parameters change over the execution of the function.

Since the problem of crash is most likely caused by a read/write concurrency on CC status attribute and since parameters are pointers, I think this is the cause.

I understand that this patch not really solves the problem, but it solves the consequences of it. And because of the problem is a concurrency one, it seems fine (to me) to just wait for 5 seconds to correctly finish the 'write' operation and read again 5s later since it's avoiding the issue.

pkg/controller/cassandracluster/reconcile.go
PERES-Richard: That's also what I thought until the operator crashed even with 1st nil check :/ (cf. previous message)

pkg/controller/cassandracluster/reconcile_test.go
PERES-Richard: Hum ok. So do think that '_' instead of 'err' it's better since this test is not made to trigger this issue ?

test/kuttl/multi-dcs/00-createCluster.yaml
PERES-Richard: Okay good to know, i'm changing it right away !

test/kuttl/multi-dcs/02-unallowedScaleDown.yaml
PERES-Richard: I wasn't able to reproduce with only 1 DC. Only with 2+ DC and 1 or more nodesPerRack

test/kuttl/multi-dcs/01-assert.yaml
PERES-Richard: Oh ok I had a problem with '| length' that was interpreting the rest of the line but with "( )" it should work. I pushing a new version right away

PERES-Richard: Seems like the problem its still there, I will keep the previous one :/

cscetbon: what problem ? I've tested and it works as expected

test/kuttl/multi-dcs/02-unallowedScaleDown.yaml
cscetbon: ok :/

pkg/controller/cassandracluster/reconcile_test.go
cscetbon: Nah. What I'm saying is that a test should never have tests like this. The conditions are given and the function must error or not. In that test it seems to depend which doesn't make any sense and should be removed/fixed

pkg/controller/cassandracluster/reconcile.go
cscetbon: it doesn't make any sense even with concurrency

cscetbon:
@PERES-Richard please test and review this PR

PERES-Richard (commented):
I've tested on GKE and all seems OK to me :)

pkg/controller/cassandracluster/deploy_cassandra.go
```diff
@@ -71,17 +71,20 @@ func (rcc *ReconcileCassandraCluster) ensureCassandraServiceMonitoring(cc *api.C
...
-
-	pdb := generatePodDisruptionBudget(cc.Name, cc.Namespace, labels, k8s.AsOwner(cc),
-		intstr.FromInt(int(cc.Spec.MaxPodUnavailable)))
+	pdb := rcc.podDisruptionBudgetEnvelope(cc)
 	err := rcc.CreateOrUpdatePodDisruptionBudget(pdb)
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		logrus.Errorf("CreateOrUpdatePodDisruptionBudget Error: %v", err)
 	}
 	return err
 }
```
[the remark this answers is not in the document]
cscetbon: It's just an envelope used to generate the PDB, hence its name

deploy/crds/db.orange.com_cassandraclusters_crd.yaml
```diff
@@ -1169,7 +1169,7 @@ spec:
                       type: array
                   type: object
                 unlockNextOperation:
-                  description: Very special Flag to hack CassKop reconcile loop - use with really good Care
+                  description: Very special Flag to hack CassKop reconcile loop - use with really good care
```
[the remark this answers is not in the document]
cscetbon: No one has to. Upgrading a CRD can be done or not depending on the change. On our side it doesn't really matter as long as in the changelog we don't specify a need to upgrade it.

PR #341, opened 2021-07-07
Document how to upgrade from v1 to v2

cscetbon:


PERES-Richard (commented):
This change includes a sidebar modification, so `sidebar.js` should be edited accordingly.

PERES-Richard (dismissed):

PR #375, opened 2021-10-27
Fix issue #374 with determining the cassandra version from image

### LINKED ISSUE #374: Casskop incorrectly determine version of image in case if image repository contains custom port
## Bug Report

If we don't explicitly specify the _serverVersion_ in CR, operator should parse this value from image name. Some deployments can contains proxy for image repository and in this case image name can looks like:
`cassandraImage: '127.0.0.1:44301/tungsten/cassandra:3.11.6'`

In this case operator can't detect version due to incorrect logic and leave _PRODUCT_VERSION_ env variable empty:
https://github.com/Orange-OpenSource/casskop/blob/2d00f1ab1950226b13309e5b633a1bf2ebcd86b6/pkg/controller/cassandracluster/generator.go#L634
it's lead to error in _config-builder_ init container.

**Environment**
* casskop version:
latest

**Possible Solution**
Change appropriate code in following way:
`
if serverVersion == "" {
		if len(image) >= 2 {
			version := strings.Split(image[len(image)-1], "-")
			serverVersion = version[0]
			if len(version) != 1 {
				serverVersion += ".0"
			}
		}
	}
`

ibumarskov:
| Q               | A
| --------------- | ---
| Bug fix?        | [+]
| New feature?    | [-]
| API breaks?     | [-]
| Deprecations?   | [-]
| Related tickets | fixes #374

### What's in this PR?
Fix cassandra version definition in case we have custom port for image repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ orbs:
           - run:
               name: Set up K3d
               command: |
-                wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash
+                wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=v4.4.8 bash
                 k3d cluster create --image ${K3S_IMAGE} --wait
                 mkdir -p ${HOME}/.kube
           - checkout:


ibumarskov:
> Please add a unit test to ensure it works as expected in 2/3 different cases.

Done, pls review

pkg/controller/cassandracluster/generator_test.go
```diff
@@ -330,6 +339,30 @@ func TestInitContainerConfigFileData(t *testing.T) {
...
+	prodVer := GetEnvVarByName(initEnvVar, "PRODUCT_VERSION")
+	assert.Equal(imageTag, prodVer.Value)
+
+	// Cassandra image contains custom port for artifactory and cassandra version
+	imageTag = "1.0.1"
+	cc.Spec.CassandraImage = fmt.Sprintf("artifactory-name:44301/repo/cassandra:%s", imageTag)
+	initEnvVar = initContainerEnvVar(cc, &cc.Status, cassieResources, dcRackName)
+	prodVer = GetEnvVarByName(initEnvVar, "PRODUCT_VERSION")
+	assert.Equal(imageTag, prodVer.Value)
+}
```
cscetbon, line 364: @ibumarskov please refactor your code as a loop instead. You can keep the same imageTag and just try both format with the same code.

cscetbon:
I got some permissions issues to trigger all our tests and have to wait to get someone on it and fix it.

pkg/controller/cassandracluster/generator_test.go
ibumarskov: Done

fdehay (approved):
seems ok for me and the integration tests work

